PR #38, opened 2021-01-15
fix(flap): set correct beg amount

e18r:
File flap.beg in the form (1 + percentage) * WAD instead of just percentage * WAD

### DIFF
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -366,12 +366,12 @@ contract DssExecLib {
     /**
         @dev Set the rate of increasing amount of MKR out for auction during debt auctions. Amount will be converted to the correct internal precision.
         @dev MKR amount is increased by this rate every "tick" (if auction duration has passed and no one has bid on the MKR)
-        @dev Equation used for conversion is (pct + 100,000) * WAD / 100,000 (ex. changes 50% to 150% WAD needed for pad)
+        @dev Equation used for conversion is (1 + pct / 10,000) * WAD
         @param _flop   The address of the Flopper core contract
         @param _pct_bps    The pct, in basis points, to set in integer form (x100). (ex. 5% = 5 * 100 = 500)
     */
     function setDebtAuctionMKRIncreaseRate(address _flop, uint256 _pct_bps) public {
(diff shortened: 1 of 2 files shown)


gbalabasquer:
The PR looks good, however it isn't the only percentage value with error. So far `mat` has the same problem.

gbalabasquer:
Same for `flop` and `flip` `beg`.

e18r:
@gbalabasquer working on it

e18r:
@gbalabasquer I will create another pull request for these

gbalabasquer:
Let's do everything here.

e18r:
I also found `flop.beg` and `flap.beg` that need to be updated.

gbalabasquer:
> I also found `flop.beg` and `flap.beg` that need to be updated.

Yeah I've mentioned them above ^^

gbalabasquer:
`flop.pad` is also a rate in the form of `1.X`. It's correctly implemented now, however in a different way than the other one. We should standardize it.

e18r:
I may be biased, but my way is better.

brianmcmichael:
We're going to route around this in the exec this week. Let's fix it all up here and make sure everything is captured by the tests, we can deploy a new version next week sometime when gas is cheaper.

I'm with @gbalabasquer no judgment on whether your way is better or not. If you want to do it your way though we should update all of the other usages so that they are standardized.

gbalabasquer:
The only advantage I see on adding up 1000 to the bps is that we need to use the unit just once.`beg` and `chop` are `wad` when `mat` is `ray`.
The other solution implies that we need to also make sure the addition is using the correct unit a part of the correct div function.

e18r:
> So far `mat` has the same problem.

`mat` is different. It is correctly specified in the form pct * RAD, instead of (1 + pct) * RAD. It's just a pretty big percentage. Also, the code makes sure it doesn't surpass 1,000 %.

So I think it's good as it is.

e18r:
@gbalabasquer @brianmcmichael ready for review

gbalabasquer:
`mat` needs to be in the form of `1.x * RAY` .

brianmcmichael (approved):
Going to merge this for now to progress with some other work. We should have these tested in a new PR I think.

PR #43, opened 2021-01-15
feat: added lerp module with tests

hexonaut:
Added the `lerp` module accompanying tests. This module is being requested more and more, so it makes sense to me to add it to exec lib.


src/DssExecLib.sol
```diff
@@ -19,6 +19,7 @@
 pragma solidity ^0.6.7;
 
 import "./MathLib.sol";
+import "./utils/Lerp.sol";
```
brianmcmichael, line 22: Rather than pulling in all of this code, consider creating a LerpFab contract, see our ["FLIP_FAB"](https://etherscan.io/address/0x4ACdbe9dd0d00b36eC2050E805012b8Fc9974f2b#code) that deploys a new lerp.

The idea here is that once it's in good shape this contract will be pretty static, so if you have `newLerp()` function on an external Factory we can reach out from here and call that. By passing in a factory address as a parameter we'd be able to modify it easily later by just pointing to the new contract.

src/DssExecLib.sol
```diff
@@ -831,4 +832,37 @@ contract DssExecLib {
...
+    /**
+        @dev Initiate linear interpolation on an administrative value over time.
+        @param _target      The target contract
+        @param _what        The target parameter to adjust
+        @param _start       The start value for the target parameter
+        @param _end         The end value for the target parameter
+        @param _duration    The duration of the interpolation
+    */
+    function linearInterpolation(address _target, bytes32 _what, uint256 _start, uint256 _end, uint256 _duration) public returns (address) {
+        Lerp lerp = new Lerp(_target, _what, _start, _end, _duration);
```
brianmcmichael, line 848: taking what I mentioned above, we'd want to pass an `address _factory` as a parameter and then we can just call something like `newLerp()` on it from here.

hexonaut:
Moved lerp out into a separate repo: https://github.com/BellwoodStudios/dss-lerp
There is now a dependency on this PR being merged: https://github.com/makerdao/dss-interfaces/pull/45

Is FLIP_FAB in the chainlog? If so, should I add a LERP_FAB as well and pull the address from there?

hexonaut:
Reopened here: https://github.com/makerdao/dss-exec-lib/pull/60 . Closing this.

PR #44, opened 2021-01-25
Cleanup in actions

brianmcmichael:
* Internalizes all functions so they won't compile in if unused.
* Removes libcall functions and uses _dcall directly to remove some LoC
* test update

### DIFF
--- a/src/DssAction.sol
+++ b/src/DssAction.sol
@@ -114,9 +114,9 @@ abstract contract DssAction {
         return ChainlogLike(LOG).getAddress(key);
     }
 
-    function _dcall(bytes memory data) internal {
+    function libcall(bytes memory data) internal {
         (bool ok,) = lib.delegatecall(data);
-        require(ok, "fail");
+        require(ok, "DssAction/failed-lib-call");
     }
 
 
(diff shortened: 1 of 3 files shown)


gbalabasquer (approved):
Changes look good and tests are passing

src/DssAction.sol
```diff
@@ -90,370 +90,293 @@ abstract contract DssAction {
...
         return _flip;
     }
 
+    function getChangelogAddress(bytes32 key) internal view returns (address) {
+        return ChainlogLike(LOG).getAddress(key);
+    }
+
     function _dcall(bytes memory data) internal {
         (bool ok,) = lib.delegatecall(data);
         require(ok, "fail");
```
e18r, line 119: "DssAction/failed-lib-call"

src/DssAction.sol
```diff
@@ -90,370 +90,293 @@ abstract contract DssAction {
...
+    function flip(bytes32 ilk) internal view returns (address) {
+        (,,,, address _flip,,,) = RegistryLike(reg()).ilkData(ilk);
         return _flip;
     }
 
+    function getChangelogAddress(bytes32 key) internal view returns (address) {
+        return ChainlogLike(LOG).getAddress(key);
+    }
+
     function _dcall(bytes memory data) internal {
```
e18r, line 117: Maybe this can now be renamed as `libCall`. In any case, I think the underscore should be removed as we don't follow that convention anywhere else.

e18r (approved):
LGTM. A couple minor comments surrounding the `_dcall` function.

Issue #41, opened 2021-01-15
Using a Solidity library


hexonaut:
I'm curious as to the design decision to use inherited delegate calls to DssExecLib as a contract instead of using it as a library. It seems like having the solidity compiler do the heavily lifting for that would be not only less code to manage, but reduce the chance of error by function signature typos. Solidity libraries can be dynamically linked as well to reduce the amount of code getting deployed into the spell.

brianmcmichael:
Solidity libraries aren't really useful for modifying state. They're commonly used for things like safemath and other pure functions that take parameters and return results. The original plan was to use native libraries but we ended up not being able to modify system params though it.

hexonaut:
Didn't realize that. How did it prevent you from doing that? Compiler error? Runtime revert?

brianmcmichael:
No revert, which was really promising until we started running tests. It basically manipulates the state of the calling contract instead of the contract you're expecting to call. Now that I'm thinking of it, this might be very dangerous if called within the context of the pause. We probably need to add some checks in the spell tests to ensure that pause values are not changed from pre- to post- test unless explicitly called.

hexonaut:
Would be interesting to look at the bytecode for the library. Are external calls in libraries using delegate calls or something instead of regular calls? Maybe there is a way to explicitly set it to use a call. I'll run some tests to see what's going on.

hexonaut:
I wrote up a proof of concept to test using libraries: https://github.com/hexonaut/lib-test

If you run `./test.sh` on Kovan it updates the LibTest contract from the library just fine. I tested both the embedded use case as well as the dynamically linked one. I think it could be worthwhile to make this switch as it will remove the need to replicate the function calls in `DssAction`. Instead of calling functions like `libCall("setIlkDebtCeiling(address,bytes32,uint256)", vat(), ilk, amount);` we could just do `ExecLib.setIlkDebtCeiling(vat(), ilk, amount)` to get compiler safety.

hexonaut:
This is done.

PR #46, opened 2021-02-03
refactor: convert manual calls into solidity library format

hexonaut:
Implementation of this idea: https://github.com/makerdao/dss-exec-lib/issues/41

This will give us compile-time errors for type/naming errors when calling out to the exec-lib. Eliminates the manual libCall(<string sig>).

Library can be used to embed code into the spell until the exec-lib is well tested. This is what the `internal` modifier means for libraries. We can then save gas by switching the modifiers on DssExecLib to `public`, deploying the library and adding a library dynamic link. Example here: https://github.com/hexonaut/lib-test/blob/master/config.json#L23

### DIFF
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -93,6 +93,16 @@ interface RegistryLike {
     function info(bytes32) external view returns (
         string memory, string memory, uint256, address, address, address, address
     );
+    function ilkData(bytes32) external view returns (
+        uint256       pos,
+        address       gem,
+        address       pip,
+        address       join,
+        address       flip,
+        uint256       dec,
+        string memory name,
+        string memory symbol
+    );
 }
 
 // https://github.com/makerdao/dss-chain-log
@@ -100,6 +110,7 @@ interface ChainlogLike {
     function setVersion(string calldata) external;
     function setIPFS(string calldata) external;
     function setSha256sum(string calldata) external;
+    function getAddress(bytes32) external view returns (address);
     function setAddress(bytes32, address) external;
     function removeAddress(bytes32) external;
 }
@@ -112,7 +123,7 @@ interface IAMLike {
 }
 
 
-contract DssExecLib {
+library DssExecLib {
 
     using MathLib for *;
 
@@ -125,7 +136,7 @@ contract DssExecLib {
         @param _key Access key for the address (e.g. "MCD_VAT")
         @param _val The address associated with the _key
     */
-    function setChangelogAddress(address _log, bytes32 _key, address _val) public {
+    function setChangelogAddress(address _log, bytes32 _key, address _val) internal {
         ChainlogLike(_log).setAddress(_key, _val);
     }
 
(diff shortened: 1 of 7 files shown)


gbalabasquer:
So I had played with this exact change in my local last week and the tests weren't passing due some weird errors. Now I'm realizing these errors happened because in the library I had kept the functions as `public`. With the `internal` everything seems to work fine.

However the idea of the `DssExecLib` is having it pre-deployed and not embed in the spell code, not only to save gas but also to reduce the amount of code for reviewing each week. So I think whatever is the final desired solution (contract or library) it should be pre-deployed from day one.

I'm not experienced at all using these solidity libraries so not really sure how we can test it as a pre-deployed with hevm. If the pre-deployed library is added at compilation time, don't think we will be able to actually test it for a library that needs to be deployed on testing time.

So we need to evaluate that even if it works as intended, the library solution doesn't add more complexity to the testing set up and deployment of each week. I can see some benefits in how the code looks, although not sure if that deserves more burden in the set up.

hexonaut:
The `public` and `internal` modifiers were what messed me up when I was testing this on my PoC repo. Basically `internal` sets the library to embed-only. `external` and `public` allow a call from a pre-deployed library however you need to specify the location of that library at compile time or weird errors start coming up as a placeholder is inserted into the bytecode instead. Don't ask me why they don't just say missing library instead lol. I'll ask the dapp tools people if it's possible to test this on a mainnet fork in the dapp test suite.

hexonaut:
After talking with the dapptools guys it appears that `dapp test` does everything we need. Running `dapp --use solc:0.6.11 test -v --rpc` works fine with the `public` visibility modifiers. `dapp test` will automatically generate libraries at pre-generated addresses then feed those into the mainnet fork as existing state. The issue is that `test.sh` is splitting out the `dapp build` from the `hevm dapp-test` commands so the state gets lost.

hexonaut:
I've updated `test.sh` to fix the library deploy + link stage. We should be good to test this next week I think.

gbalabasquer:
Oh that's a nice finding. Let's see what @brianmcmichael thinks about this change now it is working.
If we finally decide to use it, regarding spell testing and deployment, we just need to add the pre-deployed library address at compilation time, right? I guess we would do so in the Makefile.

hexonaut:
Yes this can be added as a custom "standard json" file fed to dapp build process. You can see the line added here: https://github.com/hexonaut/lib-test/blob/master/config.json#L23 which can then be supplied via `DAPP_STANDARD_JSON="config.json" dapp build`.

gbalabasquer:
Sounds good

src/DssExecLib.sol
```diff
@@ -234,7 +245,7 @@ contract DssExecLib {
     /******************************/
     // function setGlobalDebtCeiling(uint256 _amount) public { setGlobalDebtCeiling(vat(), _amount); }
     /**
-        @dev Set the global debt ceiling. Amount will be converted to the correct internal precision.
```
brianmcmichael, line 237: Was this a find/replace thing? The `internal` referred to the internal system accounting. Maybe these comments should be re-written to be clearer.

hexonaut: Yes I did a find and replace. I'll revert the comments I accidentally hit.

hexonaut:
Opened for a new local branch: https://github.com/makerdao/dss-exec-lib/pull/50

PR #47, opened 2021-02-05
File directly to correct division precision loss

brianmcmichael:
Increase and decrease functions were dividing `Line` by `RAD` which might result in some precision loss of the Global Line. This removes the division and files directly against the vat.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
-all          :; dapp --use solc:0.6.7 build
-#build		 :; SOLC_FLAGS="--libraries src/MathLib.sol:MathLib:0x987dc24Dd903F626E8C55769BFc684871Fa28E24" dapp --use solc:0.6.7 build --extract
-build		 :; dapp --use solc:0.6.7 build --extract
+all          :; dapp --use solc:0.6.11 build
+#build		 :; SOLC_FLAGS="--libraries src/MathLib.sol:MathLib:0x987dc24Dd903F626E8C55769BFc684871Fa28E24" dapp --use solc:0.6.11 build --extract
+build		 :; dapp --use solc:0.6.11 build --extract
 clean        :; dapp clean
 test         :; ./test.sh
 # KOVAN 0x987dc24Dd903F626E8C55769BFc684871Fa28E24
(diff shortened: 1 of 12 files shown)


e18r (approved):
LGTM

hexonaut (approved):
LGTM 👍

hexonaut:
Actually, you probably want to make this same change for `increaseIlkDebtCeiling` and `decreaseIlkDebtCeiling` as the same issue applies there.

e18r (approved):

gbalabasquer (approved):
LGTM, only thing I can say is IMO doing `_amount < MathLib.WAD` is a bit confusing. I know we are trying to limit the size of amount considering that needs to be multiplied by `* RAD`, so this way we can catch up users that by mistake would be sending the full number. However the unit `WAD` IMO is confusing as has nothing to do with debt ceilings. If we want to block users that are sending the full number, I think it would be better represented with:
```
_amount < MathLib.RAD
```

e18r:
I agree it can be confusing. However, `amount < RAD` doesn't prevent people from sending `wad`s, which is probably a common mistake. Your second example works better, but it's a bit arbitrary. In order to prevent the confusion, we can encapsulate the current WAD check into a function:

    require(isRegularAmount(_amount))

Maybe we should also come up with a name for this kind of amount: how about `lay`?

gbalabasquer:
Right, the `RAD` solution doesn't prevent sending `WAD` which is also a bit arbitrary as it is still a huge number.
The `100 * BILLION` is arbitrary but IMO has more to do with a common sense max debt ceiling. Today we are in the 2-3 BILLION range, so we aren't too far from that limit but with enough room to grow for a couple of years at least.

brianmcmichael:
All of the `require` statements here are merely precision checks. If we want to start enforcing actual limits on values it should be part of a separate PR and encompass the entire library.

gbalabasquer:
Fair enough, I just wondering if doing `_amount < MathLib.WAD` will get users confused about the final unit of the debt ceilings.
Maybe with a comment that says something "we are expecting a regular number (not WAD, RAY or RAD)". But we can discuss that in another PR.

PR #49, opened 2021-02-07
ExecLib Review

hexonaut:
Made some fixes after reviewing the exec-lib. This review covers `MathLib.sol`, `DssExecLib.sol` and `DssAction.sol`. I did not cover the other contracts. Proposed changes:

 * Some comment, spacing fixes.
 * Method overloads for the methods with bools.
 * Added two new constants for 100% rates and bps.
 * Replaced the upper limit of stability fees, global SF and DSR to 100% instead of `2 * RAY` as a sane upper limit. `2 * RAY` is some super high number that would be catastrophic to set.
 * MathLib constants changed to internal as we do not want to deploy them as another library.

### DIFF
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -264,7 +264,7 @@ contract DssExecLib {
         @param _rate   The accumulated rate (ex. 4% => 1000000001243680656318820312)
     */
     function setDSR(address _pot, uint256 _rate) public {
-        require((_rate >= MathLib.RAY) && (_rate < 2 * MathLib.RAY));  // "LibDssExec/dsr-out-of-bounds"
+        require((_rate >= MathLib.RAY) && (_rate <= MathLib.RATES_ONE_HUNDRED_PCT));  // "LibDssExec/dsr-out-of-bounds"
         Fileable(_pot).file("dsr", _rate);
     }
     /**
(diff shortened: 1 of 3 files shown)


src/DssExecLib.sol
```diff
@@ -264,7 +264,7 @@ contract DssExecLib {
         @param _rate   The accumulated rate (ex. 4% => 1000000001243680656318820312)
     */
     function setDSR(address _pot, uint256 _rate) public {
-        require((_rate >= MathLib.RAY) && (_rate < 2 * MathLib.RAY));  // "LibDssExec/dsr-out-of-bounds"
+        require((_rate >= MathLib.RAY) && (_rate <= MathLib.RATES_ONE_HUNDRED_PCT));  // "LibDssExec/dsr-out-of-bounds"
```
brianmcmichael, line 267: This is fine, I was wondering if we should go higher, but if someone wants to set the rate to over 100% they should probably be doing it manually anyways as I'd imagine it will be an extreme circumstance. `2*RAY` was a precision check to ensure a `RAD` didn't get passed in, but I like the tighter guard rails.

src/MathLib.sol
```diff
@@ -20,11 +20,14 @@ pragma solidity ^0.6.11;
...
-    uint256 constant public RAD      = 10 ** 45;
-    uint256 constant public THOUSAND = 10 ** 3;
-    uint256 constant public MILLION  = 10 ** 6;
+    uint256 constant internal WAD      = 10 ** 18;
+    uint256 constant internal RAY      = 10 ** 27;
+    uint256 constant internal RAD      = 10 ** 45;
+    uint256 constant internal THOUSAND = 10 ** 3;
+    uint256 constant internal MILLION  = 10 ** 6;
+
+    uint256 constant internal BPS_ONE_HUNDRED_PCT     = 10 * THOUSAND;
```
brianmcmichael, line 29: This is fine, but what about adding `BPS_ONE_PCT` above this and then on this line doing `100 * BPS_ONE_PCT` here? That will also give people access to `MathLib.BPS_ONE_PCT` elsewhere.

src/DssExecLib.sol
hexonaut: Yeah my thinking was the same for > 100%. Everything is still do-able outside of DssExecLib, so let's optimize for the common path.

PR #50, opened 2021-02-11
refactor: convert manual calls into solidity library format

hexonaut:
Re-opened from here so it's on a local branch: https://github.com/makerdao/dss-exec-lib/pull/46

Implementation of this idea: #41

This will give us compile-time errors for type/naming errors when calling out to the exec-lib. Eliminates the manual libCall().

Library can be used to embed code into the spell until the exec-lib is well tested. This is what the internal modifier means for libraries. We can then save gas by switching the modifiers on DssExecLib to public, deploying the library and adding a library dynamic link. Example here: https://github.com/hexonaut/lib-test/blob/master/config.json#L23

### DIFF
--- a/src/DssAction.sol
+++ b/src/DssAction.sol
@@ -20,48 +20,20 @@
 pragma solidity ^0.6.11;
 
 import "./CollateralOpts.sol";
-
-// https://github.com/makerdao/dss-chain-log
-interface ChainlogLike {
-    function getAddress(bytes32) external view returns (address);
-}
-
-interface RegistryLike {
-    function ilkData(bytes32) external view returns (
-        uint256       pos,
-        address       gem,
-        address       pip,
-        address       join,
-        address       flip,
-        uint256       dec,
-        string memory name,
-        string memory symbol
-    );
-}
-
-// Includes Median and OSM functions
-interface OracleLike {
-    function src() external view returns (address);
-    function lift(address[] calldata) external;
-    function drop(address[] calldata) external;
-    function setBar(uint256) external;
-    function kiss(address) external;
-    function diss(address) external;
-    function kiss(address[] calldata) external;
-    function diss(address[] calldata) external;
-}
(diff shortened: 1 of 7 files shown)


src/DssAction.sol
```diff
@@ -113,281 +84,276 @@ abstract contract DssAction {
...
-        libcall(abi.encodeWithSignature("setContract(address,bytes32,bytes32,address)", base, ilk, what, addr));
+        DssExecLib.setContract(base, ilk, what, addr);
     }
 
     /******************************/
     /*** System Risk Parameters ***/
     /******************************/
     function setGlobalDebtCeiling(uint256 amount) internal {
-        libcall(abi.encodeWithSignature("setGlobalDebtCeiling(address,uint256)", vat(), amount));
+        DssExecLib.setGlobalDebtCeiling(vat(), amount);
```
gbalabasquer, line 151: I think we discussed it with @brianmcmichael at some point but don't remember exactly if there wasn't a strong reason to not do it, but if we could define `vat()` and all the other contracts inside the `DssExecLib`, most of these functions could be directly called without `DssActions` as in the spell action you should just change `setGlobalDebtCeiling(uint256 amount)` to `DssExecLib .setGlobalDebtCeiling(uint256 amount)`. I think that would enable the real power of using the library.

hexonaut: This would remove a lot of code duplication if it works. I've found myself devoting a good chunk of review time double checking parameters are being forwarded in the correct order, the names match, etc. The less the better.

gbalabasquer: Basically the only function that would be necessary in `DssActions` would be: `addNewCollateral(...)`.

gbalabasquer: Yeah I agree, now that the library simplified the `abi.encode` thing, we could go one step further and reduce `DssActions` to its minimum expression.

PR #51, opened 2021-02-11
Fix to work with official latest release

gbalabasquer:
`--rpc` doesn't work with latest official release of dapptools yet, only installing `master` branch.
This fix makes the script to work with the latest official release.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -4,4 +4,4 @@ set -e
 [[ "$ETH_RPC_URL" && "$(seth chain)" == "ethlive"  ]] || { echo "Please set a mainnet ETH_RPC_URL"; exit 1;  }
 
 # SOLC_FLAGS="--optimize --optimize-runs 1" dapp --use solc:0.6.11 build
-dapp --use solc:0.6.11 test -v --rpc
+dapp --use solc:0.6.11 test -v --rpc-url="$ETH_RPC_URL"


test.sh
```diff
@@ -4,4 +4,4 @@ set -e
 [[ "$ETH_RPC_URL" && "$(seth chain)" == "ethlive"  ]] || { echo "Please set a mainnet ETH_RPC_URL"; exit 1;  }
 
 # SOLC_FLAGS="--optimize --optimize-runs 1" dapp --use solc:0.6.11 build
-dapp --use solc:0.6.11 test -v --rpc
+dapp --use solc:0.6.11 test -v --rpc-url="$ETH_RPC_URL"
```
brianmcmichael, line 7: Is this forward-compatible? Can we use both flags?

gbalabasquer: It should be: https://github.com/dapphub/dapptools/blob/master/src/dapp/libexec/dapp/dapp#L27

You can choose to use `--rpc-url` and pass the url or just `--rpc` and it will set it using the environment variable.
The problem is right now the official version has broken the `--rpc` option, but both should work as different alternatives

gbalabasquer: So basically what I did is change to the non-broken alternative.

brianmcmichael: tested, it works with the newer version

PR #55, opened 2021-02-11
Remove most of the functions from DssActions and use them directly fr…

gbalabasquer:
…om the ExecLib


gbalabasquer:
Still need to see how to apply all the tests that were previously in DssActions.t.sol into tests that use the lib directly. But so far everything seems working fine.

gbalabasquer:
Ok tests were added back, just linking them directly to DssExecLib when corresponds.

PR #56, opened 2021-02-11
Remove officeHours from constructor

brianmcmichael:
Removes the need to pass officeHours through the constructor. Returns true by default. If dev wants officeHours off, they can override the function and return false.


gbalabasquer (approved):
LGTM

PR #57, opened 2021-02-12
Add mkr, dai, and pauseProxy helpers

brianmcmichael:
Minor update, adds a couple of helpers

### DIFF
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -168,7 +168,7 @@ library DssExecLib {
     /*** Core Address Helpers ***/
     /****************************/
     function dai()        public view returns (address) { return getChangelogAddress("MCD_DAI"); }
-    function mkr()        public view returns (address) { return getChangelogAddress("MCD_ADM"); }
+    function mkr()        public view returns (address) { return getChangelogAddress("MCD_GOV"); }
     function vat()        public view returns (address) { return getChangelogAddress("MCD_VAT"); }
     function cat()        public view returns (address) { return getChangelogAddress("MCD_CAT"); }
     function jug()        public view returns (address) { return getChangelogAddress("MCD_JUG"); }
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -167,6 +167,8 @@ library DssExecLib {
     /****************************/
     /*** Core Address Helpers ***/
     /****************************/
+    function dai()        public view returns (address) { return getChangelogAddress("MCD_DAI"); }
+    function mkr()        public view returns (address) { return getChangelogAddress("MCD_ADM"); }
     function vat()        public view returns (address) { return getChangelogAddress("MCD_VAT"); }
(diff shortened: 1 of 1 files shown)


src/DssExecLib.sol
```diff
@@ -167,6 +167,8 @@ library DssExecLib {
     /****************************/
     /*** Core Address Helpers ***/
     /****************************/
+    function dai()        public view returns (address) { return getChangelogAddress("MCD_DAI"); }
+    function mkr()        public view returns (address) { return getChangelogAddress("MCD_ADM"); }
```
hexonaut, line 171: Is this supposed to be pointing to DSChief? If so then maybe use `gov()` or something? `mkr()` makes me think it is pointing to the token.

src/DssExecLib.sol
```diff
@@ -180,6 +182,7 @@ library DssExecLib {
     function osmMom()     public view returns (address) { return getChangelogAddress("OSM_MOM"); }
     function govGuard()   public view returns (address) { return getChangelogAddress("GOV_GUARD"); }
     function flipperMom() public view returns (address) { return getChangelogAddress("FLIPPER_MOM"); }
+    function pauseProxy() public view returns (address) { return getChangelogAddress("MCD_PAUSE_PROXY"); }
```
hexonaut, line 185: Do we need this? `MCD_PAUSE_PROXY` should always be `address(this)`.

brianmcmichael: It will be clearer to directly set `rely`'s on the PauseProxy address

src/DssExecLib.sol
brianmcmichael: this is a bug. this should be pointing to `"MCD_GOV"`, which is the MKR token, our changelog labels are a little confusing

brianmcmichael: and it will only be `address(this)` when called in the context of the spell, these are public functions so if the library is called directly it will be something else

src/DssExecLib.sol
brianmcmichael: and it will only be address(this) when called in the context of the spell, these are public functions so if the library is called directly it will be something else

hexonaut: Readability makes sense to me, but I don't see what else will be calling this other than the pause proxy.

brianmcmichael: We used it a couple of times in [last week's spell](https://github.com/makerdao/spells-mainnet/commit/ed27cc8bebe01e65aa4f360a486eb637ce544e36#diff-c9c3998420c475f01de841ca5283ea49ce6c99080a8c6a3fdae77b097bef6778R191)

hexonaut: Yeah that was a mistake on my end. Should have just been `address(this)`.

gbalabasquer (approved):
LGTM. I don't have a strong opinion regarding adding `pauseProxy()` or not.

hexonaut (approved):
Feel free to merge with the `pauseProxy`. I'm okay either way.

PR #60, opened 2021-02-15
feat: Adding lerp support

hexonaut:
Re-opening from here: https://github.com/makerdao/dss-exec-lib/pull/43

Depends on https://github.com/makerdao/dss-interfaces/pull/45 being merged.


brianmcmichael:
Let's work on a plan to get the new version deployed and then we can update interfaces and actions to use that pattern.

src/DssExecLib.sol
```diff
@@ -764,4 +774,37 @@ library DssExecLib {
...
+    /**
+        @dev Initiate linear interpolation on an administrative value over time.
+        @param _target      The target contract
+        @param _what        The target parameter to adjust
+        @param _start       The start value for the target parameter
+        @param _end         The end value for the target parameter
+        @param _duration    The duration of the interpolation
+    */
+    function linearInterpolation(address _target, bytes32 _what, uint256 _start, uint256 _end, uint256 _duration) public returns (address) {
+        address lerp = LerpFactoryLike(lerpFab()).newLerp(_target, _what, _start, _end, _duration);
```
brianmcmichael, line 790: The current implemention of the Lerp also has a `startTime` parameter, which is missing from these fab calls. 

It's not immediately clear what the difference between the `start` and `startTime` values are, either, so we should probably document them here and in `dss-lerp`

brianmcmichael:
The dss-interfaces#45 has been merged.

src/DssExecLib.sol
```diff
@@ -120,6 +120,15 @@ interface IAMLike {
     function exec(bytes32) external returns (uint256);
 }
 
+interface LerpFactoryLike {
+    function newLerp(address target_, bytes32 what_, uint256 start_, uint256 end_, uint256 duration_) external returns (address);
```
brianmcmichael, line 124: function newLerp(bytes32 name_, address target_, bytes32 what_, uint256 startTime_, uint256 start_, uint256 end_, uint256 duration_) external auth returns (address lerp)

brianmcmichael (approved):
LGTM

talbaneth (approved):
LGTM

PR #61, opened 2021-02-16
Upgrade dependencies + set up Travis

gbalabasquer:


src/test/DssAction.t.sol
```diff
@@ -78,7 +78,7 @@ contract ActionTest is DSTest {
     FlipperMom flipperMom;
     DssAutoLine autoLine;
 
-    ChainLog log;
+    ChainLog clog;
```
brianmcmichael, line 81: lol

brianmcmichael (changes requested):
Getting an error on build here. Looks like remappings. Did you need to commit a config file?

src/test/DssAction.t.sol
gbalabasquer: :)

gbalabasquer:
> Getting an error on build here. Looks like remappings. Did you need to commit a config file?

Hmmm no, it doesn't need a config file. Which is the exact error?

gbalabasquer:
The config.json was added in another branch, which might be an issue now. I'll have to rebase and see how that affects the new dependencies.

brianmcmichael:
Tests pass locally. Remapping issue because `osm` dep was removed. `rm -rf lib/osm` did the trick.

PR #62, opened 2021-02-16
feat: Add config

hexonaut:
Switch to json standard config so we can turn on optimizations + runs in the new dapp tools format.


brianmcmichael (approved):
Looks good.

PR #63, opened 2021-02-17
Workaround to be able to compile with optimization on 0.6.11 and impo…

gbalabasquer:
…rting modules that call msize

### DIFF
--- a/allow-optimize.sh
+++ b/allow-optimize.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+set -e
+
+libFile="lib/dss/src/lib.sol"
+vatFile="lib/dss/src/vat.sol"
+medFile="lib/median/src/median.sol"
+osmMFile="lib/osm-mom/src/OsmMom.sol"
+osmFile="lib/osm-mom/lib/osm/src/osm.sol"
+originalLibContent=$(cat "$libFile")
+originalVatContent=$(cat "$vatFile")
+originalMedContent=$(cat "$medFile")
+originalOsmContent=$(cat "$osmFile")
+originalOsmMContent=$(cat "$osmMFile")
+
+function clean() {
+  echo "$originalLibContent" > "$libFile"
+  echo "$originalVatContent" > "$vatFile"
+  echo "$originalMedContent" > "$medFile"
+  echo "$originalOsmContent" > "$osmFile"
+  echo "$originalOsmMContent" > "$osmMFile"
+}
+
+trap clean EXIT
+
+content=$(sed '29,43 d' "$libFile")
+echo "$content" > "$libFile"
+
+content=$(sed '74,88 d' "$vatFile")
+echo "$content" > "$vatFile"
+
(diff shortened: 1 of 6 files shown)


test.sh
```diff
@@ -3,5 +3,13 @@ set -e
...
-dapp --use solc:0.6.11 test -v --rpc-url="$ETH_RPC_URL"
+CI=$1
+
+# shellcheck disable=SC1091
+source "./allow-optimize.sh"
+
+if [[ "$CI" == 1 ]]; then
+    dapp test -v --rpc-url="$ETH_RPC_URL"
+else
+    DAPP_SOLC_OPTIMIZE=true DAPP_BUILD_OPTIMIZE=1 DAPP_SOLC_OPTIMIZE_RUNS=200 dapp --use solc:0.6.11 test -v --rpc-url="$ETH_RPC_URL"
```
hexonaut, line 14: Change this to `DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200`

build.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -e
+
+# shellcheck disable=SC1091
+source "./allow-optimize.sh"
+
+DAPP_SOLC_OPTIMIZE=true DAPP_BUILD_OPTIMIZE=1 DAPP_SOLC_OPTIMIZE_RUNS=200 dapp --use solc:0.6.11 build
```
hexonaut, line 7: Change this to `DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=1`

test.sh
gbalabasquer: Done, set to 200 runs though

build.sh
gbalabasquer: Done, set to 200 runs though

test.sh
gbalabasquer: (1 run is just for the spell)

hexonaut (approved):
LGTM. Tested deploy on a mainnet fork with optimizations on.

allow-optimize.sh
```diff
@@ -0,0 +1,38 @@
...
+  echo "$originalLibContent" > "$libFile"
+  echo "$originalVatContent" > "$vatFile"
+  echo "$originalMedContent" > "$medFile"
+  echo "$originalOsmContent" > "$osmFile"
+  echo "$originalOsmMContent" > "$osmMFile"
+}
+
+trap clean EXIT
+
+content=$(sed '29,43 d' "$libFile")
```
brianmcmichael, line 25: This works for now, but if any of these files get updated these will break. We should consider this a temporary solution.

gbalabasquer: Yeah, this is clearly a temporary workaround. I'm planning to improve it though.
Changing the lines replacement for a regex that catches it would be a better one. But yes, without any doubt the final solution should be removing the `msize` from the dependencies.

brianmcmichael (approved):

PR #64, opened 2021-02-18
First step to decouple action from the lib implementation + renaming …

gbalabasquer:
…back to the full name

### DIFF
--- a/src/DssAction.sol
+++ b/src/DssAction.sol
@@ -20,11 +20,15 @@
 pragma solidity ^0.6.11;
 
 import "./CollateralOpts.sol";
-import { DssExecLib as lib, OracleLike } from "./DssExecLib.sol";
+import { DssExecLib } from "./DssExecLib.sol";
+
+interface OracleLike {
+    function src() external view returns (address);
+}
 
 abstract contract DssAction {
 
-    using lib for *;
+    using DssExecLib for *;
 
     // Office Hours defaults to true by default.
     //   To disable office hours, override this function and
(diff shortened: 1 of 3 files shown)


src/DssAction.sol
```diff
@@ -61,52 +65,52 @@ abstract contract DssAction {
...
+        if(!co.isLiquidatable) { DssExecLib.deauthorize(_flipperMom, co.flip); }
 
         if(co.isOSM) { // If pip == OSM
             // Allow OsmMom to access to the TOKEN OSM
-            lib.authorize(co.pip, lib.osmMom());
+            DssExecLib.authorize(co.pip, DssExecLib.osmMom());
             if (co.whitelistOSM) { // If median is src in OSM
                 // Whitelist OSM to read the Median data (only necessary if it is the first time the token is being added to an ilk)
-                lib.addReaderToMedianWhitelist(address(OracleLike(co.pip).src()), co.pip);
+                DssExecLib.addReaderToMedianWhitelist(address(OracleLike(co.pip).src()), co.pip);
```
brianmcmichael, line 81: We could probably turn this into a separate function 
```
addReaderToMedianWhitelist(address pip);
```
That reads the src. Then we don't need the `OracleLike` declaration at the top.

That's a different PR though. I'll make a ticket.

brianmcmichael (approved):
I pushed a little fix to the test. If we're doing this we really should move collateral onboarding to the library, too, otherwise we have this inconsistency in the way spell actions are invoked.

Screenshot from 2021-02-18 10-40-46

PR #66, opened 2021-02-19
Remove auth from flippermom when liquidations off

brianmcmichael:
Seems we have to auth flippermom on flip before we can deauth flip on flippermom.

### DIFF
--- a/src/DssAction.sol
+++ b/src/DssAction.sol
@@ -69,13 +69,12 @@ abstract contract DssAction {
 
         address _flipperMom = DssExecLib.flipperMom();
 
-        // Allow FlipperMom to access to the ilk Flipper
-        DssExecLib.authorize(co.flip, _flipperMom);
-
         if (!co.isLiquidatable) {
-            // Remove liquidation permission from FlipperMom and Flip
-            DssExecLib.deauthorize(_flipperMom, co.flip);
-            DssExecLib.deauthorize(co.flip, _flipperMom);
+            // Disallow Cat to kick auctions in ilk Flipper
+            DssExecLib.deauthorize(co.flip, DssExecLib.cat());
+        } else {
+            // Allow FlipperMom to access to the ilk Flipper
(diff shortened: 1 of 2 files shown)


src/DssAction.sol
```diff
@@ -67,11 +67,16 @@ abstract contract DssAction {
...
+        // Allow FlipperMom to access to the ilk Flipper
         DssExecLib.authorize(co.flip, _flipperMom);
-        // Disallow Cat to kick auctions in ilk Flipper
-        if(!co.isLiquidatable) { DssExecLib.deauthorize(_flipperMom, co.flip); }
+
+        if (!co.isLiquidatable) {
+            // Remove liquidation permission from FlipperMom and Flip
+            DssExecLib.deauthorize(_flipperMom, co.flip);
+            DssExecLib.deauthorize(co.flip, _flipperMom);
+        }
```
gbalabasquer, line 79: Other option would be:
```
if (!co.isLiquidatable) {
    // Disallow Cat to kick auctions in ilk Flipper
    DssExecLib.deauthorize(co.flip, DssExecLib.cat());
} else {
    DssExecLib.authorize(co.flip, _flipperMom);
}
```

gbalabasquer, line 79: Other option would be:
```
if (!co.isLiquidatable) {
    // Disallow Cat to kick auctions in ilk Flipper
    DssExecLib.deauthorize(co.flip, DssExecLib.cat());
} else {
    DssExecLib.authorize(co.flip, DssExecLib.flipperMom());
}
```

gbalabasquer (approved):
LGTM, just a suggestion to make it a bit more efficient (not that explicit though)

src/DssAction.sol
brianmcmichael: I like it.

hexonaut (approved):
LGTM 👍

gbalabasquer:
New changes look good to me.

PR #72, opened 2021-03-18
Move COB from DssAction to DssExecLib

brianmcmichael:
* Moves collateral onboarding from the action to the lib
* Updates some tests to not rely on mainnet state

Still wondering whether we should also import `CollateralOpts` into DssAction so that it's available to the spellcrafter, otherwise they will need to do a separate explicit import.

### DIFF
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -17,6 +17,10 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 pragma solidity ^0.6.11;
+pragma experimental ABIEncoderV2;
+
+import { CollateralOpts } from "./CollateralOpts.sol";
+
 
 interface Initializable {
     function init(bytes32) external;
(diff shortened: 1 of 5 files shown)


src/DssExecLib.sol
```diff
@@ -790,6 +792,60 @@ library DssExecLib {
         RegistryLike(reg()).add(_join);
     }
 
+    // Complete collateral onboarding logic.
+    function addNewCollateral(CollateralOpts memory co) internal {
```
hexonaut, line 796: Visibility should be `public` to allow non-embedded access.

hexonaut:
Looks good to merge after the visibility fix.

src/DssExecLib.sol
brianmcmichael: Good catch. Tests should probably be updated to catch this, however, this change breaks the API since the library can't accept a struct from an external source.

I'll have to add a function back in to `DssAction` to take the struct and normalize to standard parameters. I was hoping we wouldn't have to go this route but it will probably save a lot of gas.

hexonaut: Can this be fixed with the ABIEncoderV2 ?

hexonaut:
I think I prefer using the ABIEncoderV2 vs flattening into a bunch of arrays. Is there an issue with adding the V2 encoder pragma to enable struct serialization?

src/DssExecLib.sol
gbalabasquer: I think it would be fixed and will actually make it much clearer.
At some point I had heard it added quite a lot overhead provoking more gas consumption, but we might want to test it.

brianmcmichael:
Yeah getting `UnimplementedFeatureError` out of `solc` when compiling with parameterized struct in `ABIEncoderV2`.

Looks something like:

```
./test.sh
 dapp-build: building with linked libraries
UnimplementedFeatureError: Encoding type "struct CollateralOpts memory" not yet implemented.

dapp-test: rpc block: latest
hevm: (^?!): empty Fold
CallStack (from HasCallStack):
  error, called at src/Control/Lens/Fold.hs:1310:28 in lens-4.18.1-9qmhqm9T86YJHOPOEopx4d:Control.Lens.Fold
  ^?!, called at src/EVM/Solidity.hs:342:20 in hevm-0.44.1-LcTWU2RNn7iMuZpzDnXLB:EVM.Solidity
Makefile:3: recipe for target 'test' failed
make: *** [test] Error 1
```

Haven't tested with `solc:0.7.X` but kind of wondering if it got straightened out by then.

gbalabasquer:
> Yeah getting `UnimplementedFeatureError` out of `solc` when compiling with parameterized struct in `ABIEncoderV2`.
> 
> Looks something like:
> 
> ```
> ./test.sh
>  dapp-build: building with linked libraries
> UnimplementedFeatureError: Encoding type "struct CollateralOpts memory" not yet implemented.
> 
> dapp-test: rpc block: latest
> hevm: (^?!): empty Fold
> CallStack (from HasCallStack):
>   error, called at src/Control/Lens/Fold.hs:1310:28 in lens-4.18.1-9qmhqm9T86YJHOPOEopx4d:Control.Lens.Fold
>   ^?!, called at src/EVM/Solidity.hs:342:20 in hevm-0.44.1-LcTWU2RNn7iMuZpzDnXLB:EVM.Solidity
> Makefile:3: recipe for target 'test' failed
> make: *** [test] Error 1
> ```
> 
> Haven't tested with `solc:0.7.X` but kind of wondering if it got straightened out by then.

I was playing with it for a while and having the same error. It gets solved if you declare the experimental pragma also in the caller.
That would mean the spell action itself should have the declaration as well. That's something to evaluate.
Here a PR with the functionality working: https://github.com/makerdao/dss-exec-lib/pull/75

brianmcmichael:
> I was playing with it for a while and having the same error. It gets solved if you declare the experimental pragma also in the caller.
> That would mean the spell action itself should have the declaration as well. That's something to evaluate.
> Here a PR with the functionality working: #75


This may have been the problem. I hadn't included it in the test files, which after looking at your PR must have been the issue.

hexonaut:
LGTM after https://github.com/makerdao/dss-exec-lib/pull/75 is merged in.

PR #74, opened 2021-03-19
Refactor nextCastTime to DssExecLib

brianmcmichael:
There's a lot of logic that is going in to the DssExec every week for the officehours functionality.

This creates two `pure` functions with optimized calldata that can be called directly from DssAction.

This should reduce the bytecode size of a weekly deployment with minimal impact on cast cost.

- `canCast`: returns true if the spell can be cast
- `nextCastTime`: used to determine the next castable time.


src/test/DssAction.t.sol
```diff
@@ -303,6 +303,37 @@ contract ActionTest is DSTest {
...
+    function test_nextCastTime() public {
+        assertEq(action.nextCastTime_test(1616169600, 1616169600, true), 1616169600);
+        assertEq(action.nextCastTime_test(1616169600, 1616169600, false), 1616169600);
+
+        assertEq(action.nextCastTime_test(1616256000, 1616256000, true), 1616421600);
+        assertEq(action.nextCastTime_test(1616256000, 1616256000, false), 1616256000);
+    }
+
+    function testFail_nextCastTime_eta_zero() public {
+        assertEq(action.nextCastTime_test(0, 1616256000, false), 1616256000);
```
hexonaut, line 330: Nit: Don't really need asserts here.

src/test/DssAction.t.sol
```diff
@@ -303,6 +303,37 @@ contract ActionTest is DSTest {
...
+    }
+
+    function testFail_nextCastTime_ts_zero() public {
+        assertEq(action.nextCastTime_test(1616256000, 0, false), 1616256000);
```
hexonaut, line 334: Nit: Don't really need asserts here.

hexonaut (changes requested):
Like the overall idea. A few changes suggested.

src/DssExecLib.sol
```diff
@@ -169,6 +168,61 @@ library DssExecLib {
...
+    /**
+        @dev Returns true if a time is within office hours range
+        @param _ts           The timestamp to check, usually block.timestamp
+        @param _officeHours  true if office hours is enabled.
+        @return              true if time is in castable range
+    */
+    function canCast(uint40 _ts, bool _officeHours) public pure returns (bool) {
+        if (_officeHours) {
+            uint day = (_ts / 1 days + 3) % 7;
+            require(day < 5, "DssExecLib/Can only be cast on a weekday");
```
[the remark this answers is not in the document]
gbalabasquer: I agree.

gbalabasquer:
Did we check how close to the bytecode max size we got in `DssExecLib` with this change?

src/DssExecLib.sol
```diff
@@ -169,6 +168,61 @@ library DssExecLib {
...
+            uint hour = _ts / 1 hours % 24;
+            require(hour >= 14 && hour < 21, "DssExecLib/Outside office hours");
```
gbalabasquer, line 186: Same here.

src/DssExecLib.sol
```diff
@@ -169,6 +168,61 @@ library DssExecLib {
...
+    }
+
+    /**
+        @dev Calculate the next available cast time in epoch seconds
+        @param _eta          The scheduled time of the spell plus the pause delay
+        @param _ts           The current timestamp, usually block.timestamp
+        @param _officeHours  true if office hours is enabled.
+        @return castTime     The next available cast timestamp
+    */
+    function nextCastTime(uint40 _eta, uint40 _ts, bool _officeHours) public pure returns (uint256 castTime) {
```
gbalabasquer, line 198: Do we really need the `_ts` via param? Can't we just use `block.timestamp` here?

hexonaut: I could see this being useful in the case where you need to find the cast time from some arbitrary time instead of the current time.

src/DssExecLib.sol
```diff
@@ -169,6 +168,61 @@ library DssExecLib {
...
+    /*** OfficeHours Management ***/
+    /******************************/
+
+    /**
+        @dev Returns true if a time is within office hours range
+        @param _ts           The timestamp to check, usually block.timestamp
+        @param _officeHours  true if office hours is enabled.
+        @return              true if time is in castable range
+    */
+    function canCast(uint40 _ts, bool _officeHours) public pure returns (bool) {
```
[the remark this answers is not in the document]
hexonaut: Similarly here, especially if there are multiple spells that need to be executed one after the other for testing.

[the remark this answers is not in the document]
gbalabasquer: I think for this specific case we could just cast the first one, do `hevm.warp()`, cast the next one, etc.
I just feel that is more natural to use `block.timestamp` directly than having to pass it via param. Not strongly opposed to maintain it if you guys feel is better though.

src/DssExecLib.sol
gbalabasquer: I actually have some trouble considering this function should be in the exec-lib at all.
This is not used for the spell, this is for UIs or external contracts and it is needing a nested chain of external calls to expose it. This function could be addressed as well in an external contract that receives the spell as param which has nothing to do with this repo.

gbalabasquer: Removing it would also save bytecode space.

src/DssExecLib.sol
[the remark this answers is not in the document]
brianmcmichael: If we want this to be a `pure` function it needs to be here, otherwise it needs to be a `view`

src/DssExecLib.sol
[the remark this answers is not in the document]
brianmcmichael: Sounds good. We're probably not going to lose the detailed message if we don't revert in the function, but I think it's fine.

src/DssExecLib.sol
brianmcmichael: Yes, having this in a separate contract is something that would make sense. It's bureaucratically simpler to keep it here for now until we can find somewhere to put it.

An early design had this function on the factory contract and it just called back to it.

src/DssExecLib.sol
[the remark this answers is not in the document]
gbalabasquer: Yeah, not sure which other types of usage this function will receive, I just feel for the current usage could be simpler. But up to you guys whatever prefer.

src/DssExecLib.sol
gbalabasquer: Might be just a support contract that handles it and nothing else.

hexonaut (approved):
Tests passing. LGTM. I don't have strong feelings about including the `block.timestamp` or not.

src/DssAction.sol
```diff
@@ -47,15 +53,9 @@ abstract contract DssAction {
...
-        if (officeHours()) {
-            uint day = (block.timestamp / 1 days + 3) % 7;
-            require(day < 5, "Can only be cast on a weekday");
-            uint hour = block.timestamp / 1 hours % 24;
-            require(hour >= 14 && hour < 21, "Outside office hours");
-        }
-        _;
+    // Returns the next available cast time
+    function nextCastTime(uint256 eta) external returns (uint256 castTime) {
+        castTime = DssExecLib.nextCastTime(uint40(eta), uint40(block.timestamp), officeHours());
```
brianmcmichael, line 58: Actually this needs the `require` check again because the param is `uint256` and this can silently overflow.

src/DssExecLib.sol
```diff
@@ -143,32 +143,86 @@ library DssExecLib {
...
+    /**
+        @dev Calculate the next available cast time in epoch seconds
+        @param _eta          The scheduled time of the spell plus the pause delay
+        @param _ts           The current timestamp, usually block.timestamp
+        @param _officeHours  true if office hours is enabled.
+        @return castTime     The next available cast timestamp
+    */
+    function nextCastTime(uint40 _eta, uint40 _ts, bool _officeHours) public pure returns (uint256 castTime) {
+        require(_eta != 0 && _eta < uint40(-1));  // "DssExecLib/invalid eta"
+        require(_ts  != 0 && _ts  < uint40(-1));  // "DssExecLib/invalid ts"
```
brianmcmichael, line 200: These `uint40` checks are unnecessary because the params are `uint40`

hexonaut:
I don't think we should be pulling these out into a separate library. This is going to require us to add a new library reference on every spell which takes up time and additional bytecode making reference to an additional library. I think as a practical matter I would prefer to keep everything in dss-exec-lib until we start bumping up against the bytecode limit.

gbalabasquer:
I agree with Sam. I would remove `nextCastTime` from the execlib to move it to somewhere else which the spell itself has nothing to do. Like a support contract. I know that would change things for UI people, but shouldn't be a big deal.
If that doesn't happen, then I prefer to keep one library until we reach the limit.

hexonaut:
Yeah that's a good point. It isn't used by the spell directly. It should probably be just some utility contract that can be pulled in when needed.

brianmcmichael:
I moved everything back to the library for now. I've always thought that the time stuff would be a good candidate for an external reference contract, but that's going to be a more involved deployment process. We've still got space here for now.

hexonaut (approved):
LGTM. 👍

godsflaw (approved):
LGTM

PR #75, opened 2021-03-24
Use pragma abiencoderv2 to avoid packing in the action to add a colla…

gbalabasquer:
…teral

### DIFF
--- a/src/DssAction.sol
+++ b/src/DssAction.sol
@@ -57,34 +57,4 @@ abstract contract DssAction {
         }
         _;
     }
-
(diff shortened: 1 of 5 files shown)


hexonaut (approved):
Tests passing. I like this approach a lot better than the flattened arrays.

brianmcmichael (approved):
Will close #72 in favor of this approach. There are some test innovations I'll put into a new PR.

gbalabasquer:
I actually made it to merge to #72, so don't close that one.

PR #78, opened 2021-04-27
Liquidations 2.0

brianmcmichael:
Pretty substantial overhaul.

Updates the DssExecLib for Liquidations 2.0.
Adds new functionality for Clippers, Dogs, and Calcs.

This is a destructive change that removes our ability to deploy the old "Flipper" type collaterals. If we do need to deploy a new collateral as a Flipper type, we can use [Release 0.0.6](https://github.com/makerdao/dss-exec-lib/releases/tag/v0.0.6) for the time being. If we get in the habit of continuing to do the old type we can revisit this, but I'm under the assumption that all new collaterals will be Liq2.0.

### DIFF
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -819,11 +819,10 @@ library DssExecLib {
         @param _oracle  An OSM or LP oracle contract
     */
     function whitelistOracleMedians(address _oracle) public {
-        address median0;
         (bool ok, bytes memory data) = _oracle.call(abi.encodeWithSignature("orb0()"));
         if (ok) {
             // Token is an LP oracle
-            median0 = abi.decode(data, (address));
+            address median0 = abi.decode(data, (address));
             addReaderToMedianWhitelist(median0, _oracle);
             addReaderToMedianWhitelist(OracleLike(_oracle).orb1(), _oracle);
         } else {
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -512,7 +512,7 @@ library DssExecLib {
         @param _amount The amount to set in DAI (ex. 250,000 DAI amount == 250000)
     */
     function setMaxTotalDAILiquidationAmount(uint256 _amount) public {
-        require(_amount < WAD);  // "LibDssExec/incorrect-vow-dump-precision"
+        require(_amount < WAD);  // "LibDssExec/incorrect-dog-Hole-precision"
         Fileable(dog()).file("Hole", _amount * RAD);
     }
     /**
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -1038,9 +1038,8 @@ library DssExecLib {
         // Set the ilk min collateralization ratio
         setIlkLiquidationRatio(co.ilk, co.liquidationRatio);
 
-        // Set liquidation price tolerance to 50% by default
-        // Call this again after the collateral is added to override.
-        setLiquidationBreakerPriceTolerance(co.clip, 5000);
+        // Set the price tolerance in the liquidation circuit breaker
+        setLiquidationBreakerPriceTolerance(co.clip, co.breakerTolerance);
 
         // Set a flat rate for the keeper reward
         setKeeperIncentiveFlatRate(co.ilk, co.kprFlatReward);
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -681,16 +681,24 @@ library DssExecLib {
 
     /**
         @dev Set the amout of time before an auction resets.
-        @param _ilk   The ilk to update (ex. bytes32("ETH-A"))
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
         @param _duration Amount of time before auction resets (in seconds).
     */
     function setAuctionTimeBeforeReset(bytes32 _ilk, uint256 _duration) public {
         Fileable(clip(_ilk)).file("tail", _duration);
     }
-    // clip.tail
-    //Fileable(MCD_CLIP_YFI_A).file("tail", 140 minutes);
-    // clip.cusp
-    //Fileable(MCD_CLIP_YFI_A).file("cusp", 40 * RAY / 100);
+
+    /**
+        @dev Percentage drop permitted before auction reset
+        @param _ilk     The ilk to update (ex. bytes32("ETH-A"))
+        @param _pct_bps The pct, in basis points, of drop to permit (x100).
+    */
+    function setAuctionPermittedDrop(bytes32 _ilk, uint256 _pct_bps) public {
+        require(_pct_bps < BPS_ONE_HUNDRED_PCT);
+        Fileable(clip(_ilk)).file("cusp", _pct_bps * RAY / 10000);
+    }
+
+
     // clip.chip
     //Fileable(MCD_CLIP_YFI_A).file("chip", 1 * WAD / 1000);
     // clip.tip
@@ -975,9 +983,13 @@ library DssExecLib {
         // Set the ilk time max time between bids
         // FIXME
         // setIlkBidDuration(co.ilk, co.bidDuration);
-        // Set the ilk max auction duration
-        // FIXME
-        //setIlkAuctionDuration(co.ilk, co.auctionDuration);
+
+        // Set the amount of time before an auction resets.
+        setAuctionTimeBeforeReset(co.ilk, co.auctionDuration);
+
+        // Set the allowed auction drop percentage
+        setAuctionPermittedDrop(co.ilk, co.permittedDrop);
+
         // Set the ilk min collateralization ratio
         setIlkLiquidationRatio(co.ilk, co.liquidationRatio);
 
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -374,6 +374,15 @@ library DssExecLib {
     function setContract(address _base, bytes32 _ilk, bytes32 _what, address _addr) public {
         Fileable(_base).file(_ilk, _what, _addr);
     }
+    /**
+        @dev Set a value in a contract, via a governance authorized File pattern.
+        @param _base   The address of the contract where the new contract address will be filed
+        @param _what   Name of tag for the value (e.x. "Line")
+        @param _amt    The value to set or update
+    */
+    function setValue(address _base, bytes32 _what, uint256 _amt) public {
+        Fileable(_base).file(_what, _amt);
+    }
 
     /******************************/
     /*** System Risk Parameters ***/
@@ -385,7 +394,7 @@ library DssExecLib {
     */
     function setGlobalDebtCeiling(uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-global-Line-precision"
-        Fileable(vat()).file("Line", _amount * RAD);
+        setValue(vat(), "Line", _amount * RAD);
     }
     /**
         @dev Increase the global debt ceiling by a specific amount. Amount will be converted to the correct internal precision.
@@ -394,7 +403,7 @@ library DssExecLib {
     function increaseGlobalDebtCeiling(uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-Line-increase-precision"
         address _vat = vat();
-        Fileable(_vat).file("Line", add(DssVat(_vat).Line(), _amount * RAD));
+        setValue(_vat, "Line", add(DssVat(_vat).Line(), _amount * RAD));
     }
     /**
         @dev Decrease the global debt ceiling by a specific amount. Amount will be converted to the correct internal precision.
@@ -403,7 +412,7 @@ library DssExecLib {
     function decreaseGlobalDebtCeiling(uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-Line-decrease-precision"
         address _vat = vat();
-        Fileable(_vat).file("Line", sub(DssVat(_vat).Line(), _amount * RAD));
+        setValue(_vat, "Line", sub(DssVat(_vat).Line(), _amount * RAD));
     }
     /**
         @dev Set the Dai Savings Rate. See: docs/rates.txt
@@ -413,7 +422,7 @@ library DssExecLib {
     function setDSR(uint256 _rate, bool _doDrip) public {
         require((_rate >= RAY) && (_rate <= RATES_ONE_HUNDRED_PCT));  // "LibDssExec/dsr-out-of-bounds"
         if (_doDrip) Drippable(pot()).drip();
-        Fileable(pot()).file("dsr", _rate);
+        setValue(pot(), "dsr", _rate);
     }
     /**
         @dev Set the DAI amount for system surplus auctions. Amount will be converted to the correct internal precision.
@@ -421,7 +430,7 @@ library DssExecLib {
     */
     function setSurplusAuctionAmount(uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-vow-bump-precision"
-        Fileable(vow()).file("bump", _amount * RAD);
+        setValue(vow(), "bump", _amount * RAD);
     }
     /**
         @dev Set the DAI amount for system surplus buffer, must be exceeded before surplus auctions start. Amount will be converted to the correct internal precision.
@@ -516,6 +525,14 @@ library DssExecLib {
     function setMaxTotalDAILiquidationAmount(uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-dog-Hole-precision"
         Fileable(dog()).file("Hole", _amount * RAD);
+    }
+    /**
+        @dev (LIQ 1.2) Set the maximum total DAI amount that can be out for liquidation in the system at any point. Amount will be converted to the correct internal precision.
+        @param _amount The amount to set in DAI (ex. 250,000 DAI amount == 250000)
+    */
+    function setMaxTotalDAILiquidationAmountLEGACY(uint256 _amount) public {
+        require(_amount < WAD);  // "LibDssExec/incorrect-cat-box-amount"
+        Fileable(cat()).file("box", _amount * RAD);
     }
     /**
         @dev Set the duration of time that has to pass during emergency shutdown before collateral can start being claimed by DAI holders.
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -815,7 +815,7 @@ library DssExecLib {
     /*** Oracle Management ***/
     /*************************/
     /**
-        @dev Allows an oracle to read prices from it's source feeds
+        @dev Allows an oracle to read prices from its source feeds
         @param _oracle  An OSM or LP oracle contract
     */
     function whitelistOracleMedians(address _oracle) public {
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -982,7 +982,7 @@ library DssExecLib {
         // Set the amount of time before an auction resets.
         setAuctionTimeBeforeReset(co.ilk, co.auctionDuration);
 
-        // Set the allowed auction drop percentage
+        // Set the allowed auction drop percentage before reset
         setAuctionPermittedDrop(co.ilk, co.permittedDrop);
 
         // Set the ilk min collateralization ratio
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -679,7 +679,14 @@ library DssExecLib {
         Fileable(clip(_ilk)).file("buf", _pct_bps * RAY / 10000);
     }
 
-    // FIXME
+    /**
+        @dev Set the amout of time before an auction resets.
+        @param _ilk   The ilk to update (ex. bytes32("ETH-A"))
+        @param _duration Amount of time before auction resets (in seconds).
+    */
+    function setAuctionTimeBeforeReset(bytes32 _ilk, uint256 _duration) public {
+        Fileable(clip(_ilk)).file("tail", _duration);
+    }
     // clip.tail
     //Fileable(MCD_CLIP_YFI_A).file("tail", 140 minutes);
     // clip.cusp
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -85,6 +85,8 @@ interface OracleLike {
     function diss(address) external;
     function kiss(address[] calldata) external;
     function diss(address[] calldata) external;
+    function orb0() external view returns (address);
+    function orb1() external view returns (address);
 }
 
 interface MomLike {
@@ -787,6 +789,24 @@ library DssExecLib {
     /*** Oracle Management ***/
     /*************************/
     /**
+        @dev Allows an oracle to read prices from it's source feeds
+        @param _oracle  An OSM or LP oracle contract
+    */
+    function whitelistOracle(address _oracle) public {
+        address median;
+        (bool ok, bytes memory data) = _oracle.call(abi.encodeWithSignature("orb0()"));
+        if (ok) {
+            // Token is an LP oracle
+            median = abi.decode(data, (address));
+            addReaderToMedianWhitelist(median, _oracle);
+            addReaderToMedianWhitelist(OracleLike(_oracle).orb1(), _oracle);
+        } else {
+            // Standard OSM
+            addReaderToMedianWhitelist(OracleLike(_oracle).src(), _oracle);
+        }
+    }
+
+    /**
         @dev Adds oracle feeds to the Median's writer whitelist, allowing the feeds to write prices.
         @param _median Median core contract address
         @param _feeds      Array of oracle feed addresses to add to whitelist
@@ -952,12 +972,14 @@ library DssExecLib {
             authorize(co.pip, osmMom());
             if (co.whitelistOSM) { // If median is src in OSM
                 // Whitelist OSM to read the Median data (only necessary if it is the first time the token is being added to an ilk)
-                addReaderToMedianWhitelist(address(OracleLike(co.pip).src()), co.pip);
+                whitelistOracle(co.pip);
             }
             // Whitelist Spotter to read the OSM data (only necessary if it is the first time the token is being added to an ilk)
             addReaderToOSMWhitelist(co.pip, spotter());
             // Whitelist Clipper on pip
             addReaderToOSMWhitelist(co.pip, co.clip);
+            // Allow the clippermom to access the feed
+            addReaderToOSMWhitelist(co.pip, clipperMom());
             // Whitelist End to read the OSM data (only necessary if it is the first time the token is being added to an ilk)
             addReaderToOSMWhitelist(co.pip, end());
             // Set TOKEN OSM in the OsmMom for new ilk
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -509,10 +509,11 @@ library DssExecLib {
         @dev Set the maximum total DAI amount that can be out for liquidation in the system at any point. Amount will be converted to the correct internal precision.
         @param _amount The amount to set in DAI (ex. 250,000 DAI amount == 250000)
     */
-    function setMaxTotalDAILiquidationAmount(uint256 _amount) public {
-        require(_amount < WAD);  // "LibDssExec/incorrect-vow-dump-precision"
-        Fileable(cat()).file("box", _amount * RAD);
-    }
+    //FIXME
+    //function setMaxTotalDAILiquidationAmount(uint256 _amount) public {
+    //    require(_amount < WAD);  // "LibDssExec/incorrect-vow-dump-precision"
+    //    Fileable(cat()).file("box", _amount * RAD);
+    //}
     /**
         @dev Set the duration of time that has to pass during emergency shutdown before collateral can start being claimed by DAI holders.
         @param _duration Time in seconds to set for ES processing time
@@ -631,7 +632,7 @@ library DssExecLib {
     */
     function setIlkLiquidationPenalty(bytes32 _ilk, uint256 _pct_bps) public {
         require(_pct_bps < BPS_ONE_HUNDRED_PCT);  // "LibDssExec/incorrect-ilk-chop-precision"
-        Fileable(cat()).file(_ilk, "chop", add(WAD, wdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)));
+        Fileable(dog()).file(_ilk, "chop", add(WAD, wdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)));
     }
     /**
         @dev Set max DAI amount for liquidation per vault for collateral. Amount will be converted to the correct internal precision.
@@ -640,7 +641,7 @@ library DssExecLib {
     */
     function setIlkMaxLiquidationAmount(bytes32 _ilk, uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-ilk-dunk-precision"
-        Fileable(cat()).file(_ilk, "dunk", _amount * RAD);
+        Fileable(dog()).file(_ilk, "hole", _amount * RAD);
     }
     /**
         @dev Set a collateral liquidation ratio. Amount will be converted to the correct internal precision.
@@ -659,7 +660,6 @@ library DssExecLib {
         @param _ilk   The ilk to update (ex. bytes32("ETH-A"))
         @param _pct_bps    The pct, in basis points, to set in integer form (x100). (ex. 5% = 5 * 100 = 500)
     */
-    // TODO
     function setIlkMinAuctionBidIncrease(bytes32 _ilk, uint256 _pct_bps) public {
         require(_pct_bps < BPS_ONE_HUNDRED_PCT);  // "LibDssExec/incorrect-ilk-chop-precision"
         Fileable(flip(_ilk)).file("beg", add(WAD, wdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)));
@@ -834,7 +834,7 @@ library DssExecLib {
 
         // Allow ilk Join to modify Vat registry
         authorize(_vat, _join);
-		// Allow the ilk Clipper to reduce the Dog litterbox on deal()
+		// Allow the ilk Clipper to reduce the Dog hole on deal()
         authorize(_dog, _clip);
         // Allow Dog to kick auctions in ilk Clipper
         authorize(_clip, _dog);
@@ -887,11 +887,14 @@ library DssExecLib {
         setIlkStabilityFee(co.ilk, co.ilkStabilityFee, true);
 
         // Set the ilk percentage between bids
-        setIlkMinAuctionBidIncrease(co.ilk, co.bidIncrease);
+        // FIXME
+        //setIlkMinAuctionBidIncrease(co.ilk, co.bidIncrease);
         // Set the ilk time max time between bids
-        setIlkBidDuration(co.ilk, co.bidDuration);
+        // FIXME
+        // setIlkBidDuration(co.ilk, co.bidDuration);
         // Set the ilk max auction duration
-        setIlkAuctionDuration(co.ilk, co.auctionDuration);
+        // FIXME
+        //setIlkAuctionDuration(co.ilk, co.auctionDuration);
         // Set the ilk min collateralization ratio
         setIlkLiquidationRatio(co.ilk, co.liquidationRatio);
 
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -918,6 +918,8 @@ library DssExecLib {
 
         // Allow ilk Join to modify Vat registry
         authorize(_vat, _join);
+        // Allow ilk Join to suck dai for keepers
+        authorize(_vat, _clip);
 		// Allow the ilk Clipper to reduce the Dog hole on deal()
         authorize(_dog, _clip);
         // Allow Dog to kick auctions in ilk Clipper
@@ -989,6 +991,12 @@ library DssExecLib {
         // Set liquidation price tolerance to 50% by default
         // Call this again after the collateral is added to override.
         setLiquidationBreakerPriceTolerance(co.clip, 5000);
+
+        // OPTIONAL: Set a clip.tip()  (default: 0)
+        // setKeeperIncentiveFlatRate(_ilk, _amount)
+
+        // OPTIONAL: Set a clip.chip() (default: 0)
+        // setKeeperIncentivePercent(_ilk, _pct_bps)
 
         // Update ilk spot value in Vat
         updateCollateralPrice(co.ilk);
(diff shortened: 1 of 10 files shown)


src/DssExecLib.sol
```diff
@@ -627,6 +627,7 @@ library DssExecLib {
     function setIlkMinVaultAmount(bytes32 _ilk, uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-ilk-dust-precision"
         Fileable(vat()).file(_ilk, "dust", _amount * RAD);
+        (bool ok,) = clip(_ilk).call(abi.encodeWithSignature("upchost()")); ok;
```
hexonaut, line 630: What is the purpose of this `ok;` at the end?

src/DssExecLib.sol
```diff
@@ -636,16 +637,17 @@ library DssExecLib {
     */
     function setIlkLiquidationPenalty(bytes32 _ilk, uint256 _pct_bps) public {
         require(_pct_bps < BPS_ONE_HUNDRED_PCT);  // "LibDssExec/incorrect-ilk-chop-precision"
-        Fileable(cat()).file(_ilk, "chop", add(WAD, wdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)));
+        Fileable(dog()).file(_ilk, "chop", add(WAD, wdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)));
+        (bool ok,) = clip(_ilk).call(abi.encodeWithSignature("upchost()")); ok;
```
hexonaut, line 641: Again with the `ok;` at the end?

src/DssExecLib.sol
```diff
@@ -659,31 +661,68 @@ library DssExecLib {
...
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _pct_bps  The pct, in basis points, to set in integer form (x100). (ex. 1.3x starting multiplier = 130% = 13000)
     */
-    function setIlkMinAuctionBidIncrease(bytes32 _ilk, uint256 _pct_bps) public {
-        require(_pct_bps < BPS_ONE_HUNDRED_PCT);  // "LibDssExec/incorrect-ilk-chop-precision"
-        Fileable(flip(_ilk)).file("beg", add(WAD, wdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)));
+    function setStartingPriceMultiplicativeFactor(bytes32 _ilk, uint256 _pct_bps) public {
+        require(_pct_bps < 10 * BPS_ONE_HUNDRED_PCT); // "LibDssExec/incorrect-ilk-mat-precision" // Fails if gt 10x
+        require(_pct_bps >= BPS_ONE_HUNDRED_PCT); // fail if start price is less than OSM price
+        Fileable(clip(_ilk)).file("buf", _pct_bps * RAY / 10000);
```
hexonaut, line 672: Change to `rdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)`.

src/DssExecLib.sol
```diff
@@ -659,31 +661,68 @@ library DssExecLib {
...
-        @param _duration Amount of time for auctions.
+        @dev Percentage drop permitted before auction reset
+        @param _ilk     The ilk to update (ex. bytes32("ETH-A"))
+        @param _pct_bps The pct, in basis points, of drop to permit (x100).
     */
-    function setIlkAuctionDuration(bytes32 _ilk, uint256 _duration) public {
-        Fileable(flip(_ilk)).file("tau", _duration);
+    function setAuctionPermittedDrop(bytes32 _ilk, uint256 _pct_bps) public {
+        require(_pct_bps < BPS_ONE_HUNDRED_PCT); // "LibDssExec/incorrect-clip-cusp-precision"
+        Fileable(clip(_ilk)).file("cusp", _pct_bps * RAY / 10000);
```
hexonaut, line 691: Change to `rdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)`.

src/DssExecLib.sol
```diff
@@ -659,31 +661,68 @@ library DssExecLib {
...
     }
+
+    /**
+        @dev Percentage of tab to suck from vow to incentivize keepers. Amount will be converted to the correct internal precision.
+        @param _ilk     The ilk to update (ex. bytes32("ETH-A"))
+        @param _pct_bps The pct, in basis points, of the tab to suck. (0.01% == 1)
+    */
+    function setKeeperIncentivePercent(bytes32 _ilk, uint256 _pct_bps) public {
+        require(_pct_bps < BPS_ONE_HUNDRED_PCT); // "LibDssExec/incorrect-clip-chip-precision"
+        Fileable(clip(_ilk)).file("chip", _pct_bps * WAD / 10000);
```
hexonaut, line 701: Change to `wdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)`.

src/DssExecLib.sol
```diff
@@ -659,31 +661,68 @@ library DssExecLib {
...
+    /**
+        @dev Sets the circuit breaker price tolerance in the clipper mom.
+            This is somewhat counter-intuitive,
+             to accept a 25% price drop, use a value of 75%
+        @param _clip    The clipper to set the tolerance for
+        @param _pct_bps The pct, in basis points, to set in integer form (x100). (ex. 5% = 5 * 100 = 500)
+    */
+    function setLiquidationBreakerPriceTolerance(address _clip, uint256 _pct_bps) public {
+        require(_pct_bps < BPS_ONE_HUNDRED_PCT);  // "LibDssExec/incorrect-ilk-chop-precision"
+        MomLike(clipperMom()).setPriceTolerance(_clip, _pct_bps * RAY / 10000);
```
hexonaut, line 723: Change to `rdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)`.

src/DssExecLib.sol
```diff
@@ -708,6 +747,44 @@ library DssExecLib {
...
+    /**
+        @dev Set the number of seconds for each price step.
+        @dev Abacus:StairstepExponentialDecrease only.
+        @param _calc     The address of the StairstepExponentialDecrease pricing contract
+        @param _duration Length of time between price drops [seconds]
+        @param _pct_bps Per-step multiplicative factor in basis points. (ex. 99% == 9900)
+    */
+    function initStairstepExponentialDecrease(address _calc, uint256 _duration, uint256 _pct_bps) public {
+        require(_pct_bps < 10000); // DssExecLib/cut-too-high
+        Fileable(_calc).file("cut", _pct_bps * RAY / 10000);
```
hexonaut, line 773: Change to `rdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)`.

src/DssExecLib.sol
```diff
@@ -708,6 +747,44 @@ library DssExecLib {
...
+        @dev Set the number of seconds for each price step. (99% cut = 1% price drop per step)
+             Amounts will be converted to the correct internal precision.
+        @dev Abacus:ExponentialDecrease only
+        @param _calc     The address of the ExponentialDecrease pricing contract
+        @param _pct_bps Per-step multiplicative factor in basis points. (ex. 99% == 9900)
+    */
+    function initExponentialDecrease(address _calc, uint256 _pct_bps) public {
+        require(_pct_bps < 10000); // DssExecLib/cut-too-high
+        Fileable(_calc).file("cut", _pct_bps * RAY / 10000);
+    }
```
hexonaut, line 786: I think these should be `setXXXX` instead of `initXXXX` to match the other functions. If you look at `setIlkAutoLineParameters` which is sort of an "initialization" as well. Alternatively I would rename things like `setIlkAutoLineParameters` to `initIlkAutoLineParameters`. It should be consistent whichever way you want to go.

src/DssExecLib.sol
```diff
@@ -708,6 +747,44 @@ library DssExecLib {
...
+    /**
+        @dev Set the number of seconds for each price step. (99% cut = 1% price drop per step)
+             Amounts will be converted to the correct internal precision.
+        @dev Abacus:ExponentialDecrease only
+        @param _calc     The address of the ExponentialDecrease pricing contract
+        @param _pct_bps Per-step multiplicative factor in basis points. (ex. 99% == 9900)
+    */
+    function initExponentialDecrease(address _calc, uint256 _pct_bps) public {
+        require(_pct_bps < 10000); // DssExecLib/cut-too-high
+        Fileable(_calc).file("cut", _pct_bps * RAY / 10000);
```
hexonaut, line 785: Change to `rdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)`.

src/DssExecLib.sol
```diff
@@ -802,46 +879,56 @@ library DssExecLib {
...
+        address _dog = dog();
         require(JoinLike(_join).vat() == _vat);     // "join-vat-not-match"
         require(JoinLike(_join).ilk() == _ilk);     // "join-ilk-not-match"
         require(JoinLike(_join).gem() == _gem);     // "join-gem-not-match"
         require(JoinLike(_join).dec() ==
                    ERC20(_gem).decimals());         // "join-dec-not-match"
-        require(AuctionLike(_flip).vat() == _vat);  // "flip-vat-not-match"
-        require(AuctionLike(_flip).cat() == _cat);  // "flip-cat-not-match"
-        require(AuctionLike(_flip).ilk() == _ilk);  // "flip-ilk-not-match"
+        require(AuctionLike(_clip).vat() == _vat);  // "clip-vat-not-match"
```
hexonaut, line 902: Add `require(AuctionLike(_clip).spotter() == _spotter);  // "clip-spotter-not-match"`

src/DssExecLib.sol
```diff
@@ -802,46 +879,56 @@ library DssExecLib {
...
+        require(AuctionLike(_clip).dog() == _dog);  // "clip-dog-not-match"
+        require(AuctionLike(_clip).ilk() == _ilk);  // "clip-ilk-not-match"
 
         // Set the token PIP in the Spotter
         setContract(spotter(), _ilk, "pip", _pip);
 
-        // Set the ilk Flipper in the Cat
-        setContract(_cat, _ilk, "flip", _flip);
+        // Set the ilk Clipper in the Cat
```
hexonaut, line 909: `Cat` should be `Dog`

src/DssExecLib.sol
```diff
@@ -802,46 +879,56 @@ library DssExecLib {
...
         authorize(_vat, _join);
-		// Allow the ilk Flipper to reduce the Cat litterbox on deal()
-        authorize(_cat, _flip);
-        // Allow Cat to kick auctions in ilk Flipper
-        authorize(_flip, _cat);
-        // Allow End to yank auctions in ilk Flipper
-        authorize(_flip, end());
+        // Allow ilk Join to suck dai for keepers
+        authorize(_vat, _clip);
+		// Allow the ilk Clipper to reduce the Dog hole on deal()
```
hexonaut, line 924: Tabs should be spaces.

src/DssExecLib.sol
```diff
@@ -850,14 +937,14 @@ library DssExecLib {
...
     function addNewCollateral(CollateralOpts memory co) public {
         // Add the collateral to the system.
-        addCollateralBase(co.ilk, co.gem, co.join, co.flip, co.pip);
+        addCollateralBase(co.ilk, co.gem, co.join, co.clip, co.calc, co.pip);
 
         if (co.isLiquidatable) {
-            // Allow FlipperMom to access to the ilk Flipper
-            authorize(co.flip, flipperMom());
+            // Allow clipperMom to access to the ilk Clipper
+            authorize(co.clip, clipperMom());
```
hexonaut, line 944: Need to add clipper mom access to pip as well `addReaderToOSMWhitelist(co.pip, clipperMom());`

src/DssExecLib.sol
```diff
@@ -880,23 +969,36 @@ library DssExecLib {
...
+        // Set liquidation price tolerance to 50% by default
+        // Call this again after the collateral is added to override.
+        setLiquidationBreakerPriceTolerance(co.clip, 5000);
+
+        // OPTIONAL: Set a clip.tip()  (default: 0)
+        // setKeeperIncentiveFlatRate(_ilk, _amount)
+
+        // OPTIONAL: Set a clip.chip() (default: 0)
+        // setKeeperIncentivePercent(_ilk, _pct_bps)
+
```
hexonaut, line 1001: Is there any reason in particular we are not passing these as arguments in the `CollateralOpts` struct?

hexonaut (changes requested):
Mostly looks good. I've added a few recommendations. There is a missing critical piece for the collateral addition though. Haven't looked at the tests yet, but they are passing at least.

src/DssExecLib.sol
brianmcmichael: Clears up a compiler warning. And the compilation fails if we don't use the return from the address.call()

src/DssExecLib.sol
brianmcmichael: Reduction in scope. We can always set these afterward if we intend to use them.

src/DssExecLib.sol
hexonaut: Ah ok.

src/DssExecLib.sol
hexonaut: Primoz seems to want to usually set the chip. I think the more we can make this an "all-in-one" sort of function the easier it will be to not forget something accidentally.

Even beyond this I think we should start including the autoline config params in here as well as it has become common. IMO the add collateral should be taking whatever common path we happen to be using at the time. What do you think?

src/DssExecLib.sol
```diff
@@ -178,23 +167,30 @@ library DssExecLib {
...
     function daiJoin()    public view returns (address) { return getChangelogAddress("MCD_JOIN_DAI"); }
 
+    function clip(bytes32 _ilk) public view returns (address _clip) {
+        _clip = RegistryLike(reg()).xlip(_ilk);
+    }
+
     function flip(bytes32 _ilk) public view returns (address _flip) {
-        (,,,, _flip,,,) = RegistryLike(reg()).ilkData(_ilk);
+        _flip = RegistryLike(reg()).xlip(_ilk);
     }
```
gbalabasquer, line 194: Maybe just have one `xlip` function?

src/DssExecLib.sol
gbalabasquer, line 630: Any reason why not calling directly `upchost` instead of using `call`? I don't see a way it would fail excepting overflow, which should be a situation to revert anyway.

src/DssExecLib.sol
gbalabasquer, line 641: Same here.

src/DssExecLib.sol
```diff
@@ -850,14 +937,14 @@ library DssExecLib {
...
         if (co.isLiquidatable) {
-            // Allow FlipperMom to access to the ilk Flipper
-            authorize(co.flip, flipperMom());
+            // Allow clipperMom to access to the ilk Clipper
+            authorize(co.clip, clipperMom());
         } else {
-            // Disallow Cat to kick auctions in ilk Flipper
-            deauthorize(co.flip, cat());
+            // Disallow Dog to kick auctions in ilk Clipper
+            deauthorize(co.clip, dog());
```
gbalabasquer, line 947: I wouldn't do this. For this module we have the levels. So I'd keep the authorization and do `co.clip.file("stopped", 3)`.

src/DssExecLib.sol
gbalabasquer: IMO at least `chip` should be enabled. I'd prefer both.
Regarding auto-line. Do we know if the new collaterals will come with it by default?

src/DssExecLib.sol
e18r: Soon we will only have `clip`s, so having a distinct `clip` function makes it more readable, IMO

gbalabasquer: Well we are forwarding the `ilkRegistry` function which has a `xlip` and will keep having so.
I'd prefer to be more explicit on not make the user think is bringing a `flip` when might be a `clip` or viceversa.

src/DssExecLib.sol
hexonaut: I don't see why not. Almost all current collaterals have it.

src/DssExecLib.sol
brianmcmichael: This function still works with Flips, so it would fail.

gbalabasquer: Ah right, didn't think about that. Other option would be checking the ilkRegistry class, but yeah is cheaper this way.

src/DssExecLib.sol
```diff
@@ -60,7 +60,7 @@ interface DssVat {
 
 interface AuctionLike {
     function vat() external returns (address);
-    function cat() external returns (address); // Only flip
+    function dog() external returns (address); // Only clip
     function beg() external returns (uint256);
     function pad() external returns (uint256); // Only flop
     function ttl() external returns (uint256);
```
e18r, line 66: Currently only being used for Clippers. Better change to ClipLike and remove unused functions:

    interface ClipLike {
        function vat() external returns (address);
        function dog() external returns (address);
        function ilk() external returns (bytes32);
    }

src/DssExecLib.sol
```diff
@@ -178,23 +167,30 @@ library DssExecLib {
...
     function flipperMom() public view returns (address) { return getChangelogAddress("FLIPPER_MOM"); }
+    function clipperMom() public view returns (address) { return getChangelogAddress("CLIPPER_MOM"); }
     function pauseProxy() public view returns (address) { return getChangelogAddress("MCD_PAUSE_PROXY"); }
     function autoLine()   public view returns (address) { return getChangelogAddress("MCD_IAM_AUTO_LINE"); }
     function daiJoin()    public view returns (address) { return getChangelogAddress("MCD_JOIN_DAI"); }
 
+    function clip(bytes32 _ilk) public view returns (address _clip) {
+        _clip = RegistryLike(reg()).xlip(_ilk);
+    }
+
```
e18r, line 191: maybe worth adding a `function calc(bytes32 _ilk)`?

src/DssExecLib.sol
```diff
@@ -443,14 +439,14 @@ library DssExecLib {
     }
```
e18r, line 439: update comments in lines 360 and 369

src/DssExecLib.sol
```diff
@@ -510,13 +506,17 @@ library DssExecLib {
     function setDebtAuctionMKRIncreaseRate(uint256 _pct_bps) public {
         Fileable(flop()).file("pad", add(WAD, wdiv(_pct_bps, BPS_ONE_HUNDRED_PCT)));
     }
+
+
+
+
```
e18r, line 512: remove whitespace

src/DssExecLib.sol
```diff
@@ -510,13 +506,17 @@ library DssExecLib {
...
     /**
         @dev Set the maximum total DAI amount that can be out for liquidation in the system at any point. Amount will be converted to the correct internal precision.
         @param _amount The amount to set in DAI (ex. 250,000 DAI amount == 250000)
     */
     function setMaxTotalDAILiquidationAmount(uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-vow-dump-precision"
```
e18r, line 518: update comment

src/DssExecLib.sol
brianmcmichael: what do you mean here?

src/DssExecLib.sol
```diff
@@ -510,13 +506,17 @@ library DssExecLib {
...
-        Fileable(cat()).file("box", _amount * RAD);
+        Fileable(dog()).file("Hole", _amount * RAD);
```
e18r, line 516: I know you said in the description of this PR that this is backwards-incompatible, but I see no harm in replicating this with the dog instead of replacing it. After all, we still have the cat variable.

src/DssExecLib.sol
brianmcmichael: Yeah I kind of saw this as transitional code. I don't particularily like the `xlip` nomenclature. Another option would be to put a require in here that asserts that `class` == 1 for clips and 2 for flips. Not sure.

src/DssExecLib.sol
brianmcmichael: I've added tip and chip to collateralopts. It's going to be a bigger change to convert onboarding to use dss-auto-line out of the box. I'd rather not default to it here as we can activate it with one line.

src/DssExecLib.sol
```diff
@@ -659,31 +658,69 @@ library DssExecLib {
...
-        @param _ilk   The ilk to update (ex. bytes32("ETH-A"))
-        @param _duration Amount of time for bids.
+        @dev Set the amout of time before an auction resets.
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _duration Amount of time before auction resets (in seconds).
     */
-    function setIlkBidDuration(bytes32 _ilk, uint256 _duration) public {
-        Fileable(flip(_ilk)).file("ttl", _duration);
+    function setAuctionTimeBeforeReset(bytes32 _ilk, uint256 _duration) public {
+        Fileable(clip(_ilk)).file("tail", _duration);
```
e18r, line 679: maybe make sure this doesn't exceed, say, one month? I know we're not checking duration values elsewhere, but we have to start somewhere

src/DssExecLib.sol
```diff
@@ -659,31 +658,69 @@ library DssExecLib {
...
-        @param _ilk   The ilk to update (ex. bytes32("ETH-A"))
-        @param _duration Amount of time for auctions.
+        @dev Percentage drop permitted before auction reset
+        @param _ilk     The ilk to update (ex. bytes32("ETH-A"))
+        @param _pct_bps The pct, in basis points, of drop to permit (x100).
     */
-    function setIlkAuctionDuration(bytes32 _ilk, uint256 _duration) public {
-        Fileable(flip(_ilk)).file("tau", _duration);
+    function setAuctionPermittedDrop(bytes32 _ilk, uint256 _pct_bps) public {
+        require(_pct_bps < BPS_ONE_HUNDRED_PCT); // "LibDssExec/incorrect-clip-cusp-precision"
```
e18r, line 688:
```suggestion
        require(_pct_bps < BPS_ONE_HUNDRED_PCT); // "LibDssExec/incorrect-clip-cusp-precision"
        require(_pct_bps > 0); // "LibDssExec/incorrect-clip-cusp-value"
```

src/DssExecLib.sol
e18r: some outdated comments about setting the cat contract in the vat (which has never been possible IIRC)

e18r: probably the line numbers have changed, but it's in the system configuration section

src/DssExecLib.sol
```diff
@@ -659,31 +658,69 @@ library DssExecLib {
...
+
+    /**
+        @dev Sets the circuit breaker price tolerance in the clipper mom.
+            This is somewhat counter-intuitive,
+             to accept a 25% price drop, use a value of 75%
+        @param _clip    The clipper to set the tolerance for
+        @param _pct_bps The pct, in basis points, to set in integer form (x100). (ex. 5% = 5 * 100 = 500)
+    */
+    function setLiquidationBreakerPriceTolerance(address _clip, uint256 _pct_bps) public {
+        require(_pct_bps < BPS_ONE_HUNDRED_PCT);  // "LibDssExec/incorrect-ilk-chop-precision"
```
e18r, line 720: update comment to incorrect-clippermom-price-tolerance

although I'm not sure it's worth updating these comments lol

src/DssExecLib.sol
e18r: agreed, `init` conveys the idea of initialization, but in this case we're filing parameters.

src/DssExecLib.sol
brianmcmichael: no, it's fine, let's do it right

src/DssExecLib.sol
e18r:
```suggestion
```

It's more practical to have one function to set `cut` (the below one) and one to set `step`. Update comments accordingly.

src/DssExecLib.sol
```diff
@@ -708,6 +745,44 @@ library DssExecLib {
...
+    }
+    /**
+        @dev Set the number of seconds for each price step. (99% cut = 1% price drop per step)
+             Amounts will be converted to the correct internal precision.
+        @dev Abacus:ExponentialDecrease only
+        @param _calc     The address of the ExponentialDecrease pricing contract
+        @param _pct_bps Per-step multiplicative factor in basis points. (ex. 99% == 9900)
+    */
+    function initExponentialDecrease(address _calc, uint256 _pct_bps) public {
+        require(_pct_bps < 10000); // DssExecLib/cut-too-high
```
e18r, line 782:
```suggestion
        require(_pct_bps < BPS_ONE_HUNDRED_PCT); // DssExecLib/cut-too-high
```

src/DssExecLib.sol
```diff
@@ -802,46 +877,58 @@ library DssExecLib {
...
-        authorize(_flip, end());
+        // Allow ilk Join to suck dai for keepers
+        authorize(_vat, _clip);
+        // Allow the ilk Clipper to reduce the Dog hole on deal()
+        authorize(_dog, _clip);
+        // Allow Dog to kick auctions in ilk Clipper
+        authorize(_clip, _dog);
+        // Allow End to yank auctions in ilk Clipper
+        authorize(_clip, end());
+        // Authorize the ESM to ececut in the clipper
```
e18r, line 930:
```suggestion
        // Authorize the ESM to execute in the clipper
```

src/DssExecLib.sol
```diff
@@ -850,14 +937,14 @@ library DssExecLib {
     // Complete collateral onboarding logic.
     function addNewCollateral(CollateralOpts memory co) public {
         // Add the collateral to the system.
-        addCollateralBase(co.ilk, co.gem, co.join, co.flip, co.pip);
+        addCollateralBase(co.ilk, co.gem, co.join, co.clip, co.calc, co.pip);
 
         if (co.isLiquidatable) {
-            // Allow FlipperMom to access to the ilk Flipper
-            authorize(co.flip, flipperMom());
+            // Allow clipperMom to access to the ilk Clipper
```
e18r, line 943:
```suggestion
            // Grant clipperMom access to the ilk Clipper
```

src/DssExecLib.sol
```diff
@@ -850,14 +937,14 @@ library DssExecLib {
...
+            authorize(co.clip, clipperMom());
         } else {
-            // Disallow Cat to kick auctions in ilk Flipper
-            deauthorize(co.flip, cat());
+            // Disallow Dog to kick auctions in ilk Clipper
+            Fileable(co.clip).file("stopped", 3);
```
e18r, line 947: we should have a function for this.

src/DssExecLib.sol
```diff
@@ -880,23 +969,36 @@ library DssExecLib {
         setIlkDebtCeiling(co.ilk, co.ilkDebtCeiling);
         // Set the ilk dust
         setIlkMinVaultAmount(co.ilk, co.minVaultAmount);
-        // Set the dunk size
+        // Set the hole size
         setIlkMaxLiquidationAmount(co.ilk, co.maxLiquidationAmount);
```
e18r, line 973: shouldn't the general Hole be updated as well?

e18r (approved):
LGTM, left some comments

src/DssExecLib.sol
hexonaut: Yeah for sure. Changing the auto-line should be in a different PR to be discussed separately from this.

src/DssExecLib.sol
gbalabasquer: Yes we should probably have functions for all the `file` types. But that probably should go in another PR.

src/DssExecLib.sol
gbalabasquer: The general `Hole` is an independent value of the `ilk`s. E.g. we have it set to 100M right now and it is not changing each time we add new clippers.
With the `Line` we came to this sort of rule that is the sum of all the `line`s but shouldn't be a constraint at all.

src/DssExecLib.sol
brianmcmichael: I'm not sure if this is a safe assumption. We might have stablecoin auctions with little to no drop an no expiration. Maybe let's go with 3 months for now and we can revisit if it becomes a problem.

src/DssExecLib.sol
brianmcmichael: We have `setContract` for address file types, but yeah, maybe more for values. I will go ahead and make one for this and we can expand in another PR later.

brianmcmichael: So check this out. I made a function `setValue()` to handle the files and we can update everything to use it. I've put a couple of examples in at [`86799fe` (#78)](https://github.com/makerdao/dss-exec-lib/pull/78/commits/86799fea7fa1f8a9a6ee5c8ee4436eb97cd0fd66#diff-72201ff20380f5c7fc89281be3ad2dd6bd5a992f246d41d6d9d97f71e078d40dR383)

If we like it I can update the rest of the contract to use it.

src/DssExecLib.sol
brianmcmichael: I don't really think it's more practical to have these separate in this case. Each of these calcs have unique and inconsistent fileable parameters that need to be called to properly initialize them, because they aren't handled in the constructors. Forcing them to do them both at the same time ensure they'll be properly initialized, since they aren't standard across the Abacus variations.

src/DssExecLib.sol
brianmcmichael: Here we really are initializing these though. After deploying a new calc, these component functions need to be called for it to work correctly. Ideally this would have been handled in the respective constructors but for now they effectively need to be initialized to work.

src/DssExecLib.sol
```diff
@@ -792,17 +809,17 @@ library DssExecLib {
...
             // Token is an LP oracle
-            median = abi.decode(data, (address));
-            addReaderToMedianWhitelist(median, _oracle);
+            median0 = abi.decode(data, (address));
+            addReaderToMedianWhitelist(median0, _oracle);
             addReaderToMedianWhitelist(OracleLike(_oracle).orb1(), _oracle);
         } else {
             // Standard OSM
-            addReaderToMedianWhitelist(OracleLike(_oracle).src(), _oracle);
+            addReaderToOSMWhitelist(OracleLike(_oracle).src(), _oracle);
```
hexonaut, line 822: This should be `addReaderToMedianWhitelist(OracleLike(_oracle).src(), _oracle);` as `OracleLike(_oracle).src()` is a median. The function is the same though, so it won't trip the tests.

src/DssExecLib.sol
hexonaut: That's fine. I would consider changing stuff like `setIlkAutoLineParameters` to `initIlkAutoLineParameters` as well then to convey the idea of initialization.

src/DssExecLib.sol
```diff
@@ -880,23 +1009,36 @@ library DssExecLib {
...
+
+        // Set the allowed auction drop percentage before reset
+        setAuctionPermittedDrop(co.ilk, co.permittedDrop);
+
         // Set the ilk min collateralization ratio
         setIlkLiquidationRatio(co.ilk, co.liquidationRatio);
 
+        // Set liquidation price tolerance to 50% by default
+        // Call this again after the collateral is added to override.
+        setLiquidationBreakerPriceTolerance(co.clip, 5000);
```
hexonaut, line 1034: I still think this should be explicitly specified in the collateral args just so it doesn't get accidentally overlooked.

hexonaut (commented):
Mostly looks good. A couple minor comments.

src/DssExecLib.sol
```diff
@@ -958,28 +975,29 @@ library DssExecLib {
...
         // Add the collateral to the system.
         addCollateralBase(co.ilk, co.gem, co.join, co.clip, co.calc, co.pip);
+        address clipperMom_ = clipperMom();
 
-        if (co.isLiquidatable) {
-            // Allow clipperMom to access to the ilk Clipper
-            authorize(co.clip, clipperMom());
-        } else {
+        // Grant ClipperMom access to the ilk Clipper
+        authorize(co.clip, clipperMom_);
```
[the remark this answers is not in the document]
brianmcmichael: We're handling liquidations in the clip now using the circuit breaker.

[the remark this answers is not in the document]
brianmcmichael: Yeah, you're right. I think this change came out of a conversation with @gbalabasquer and you make a good point that a governance attack could liquidate collateral out-of-pause. I will move this back and update the tests to ensure wards are `0` here. Pinging Gonza in case we're missing something.

hexonaut (approved):
LGTM 👍 I still think the `setLiquidationBreakerPriceTolerance()` should be provided as a parameter, but if you want to leave it out that's fine.

src/DssExecLib.sol
```diff
@@ -374,6 +374,15 @@ library DssExecLib {
     function setContract(address _base, bytes32 _ilk, bytes32 _what, address _addr) public {
         Fileable(_base).file(_ilk, _what, _addr);
     }
+    /**
+        @dev Set a value in a contract, via a governance authorized File pattern.
+        @param _base   The address of the contract where the new contract address will be filed
```
e18r, line 379: @param _base   The address of the contract where the new ~~contract address~~ value will be filed

src/DssExecLib.sol
e18r: Nice, I like it. We probably need more functions to handle values that aren't `uint256`.

src/DssExecLib.sol
e18r: It's true what @gbalabasquer says, but I agree it's better to have two separate functions and put `require`s on them.

src/DssExecLib.sol
e18r: Fair enough. I thought these were also going to be used for filing params in the abaci. But if that's not the main use case, then it's OK.

src/DssExecLib.sol
```diff
@@ -372,6 +374,25 @@ library DssExecLib {
...
+        @param _what   Name of tag for the value (e.x. "Line")
+        @param _amt    The value to set or update
+    */
+    function setValue(address _base, bytes32 _what, uint256 _amt) public {
+        Fileable(_base).file(_what, _amt);
+    }
+    /**
+        @dev Set a value in a contract, via a governance authorized File pattern.
+        @param _base   The address of the contract where the new contract address will be filed
```
e18r, line 388: ~~contract address~~ value

src/DssExecLib.sol
```diff
@@ -372,6 +374,25 @@ library DssExecLib {
...
+        @param _what   Name of tag for the value (e.x. "Line")
+        @param _amt    The value to set or update
+    */
+    function setValue(address _base, bytes32 _what, uint256 _amt) public {
+        Fileable(_base).file(_what, _amt);
+    }
+    /**
+        @dev Set a value in a contract, via a governance authorized File pattern.
```
e18r, line 387: ~~a value~~ an ilk-specific value

src/DssExecLib.sol
```diff
@@ -779,64 +807,82 @@ library DssExecLib {
...
-        Fileable(_calc).file("cut", _pct_bps * RAY / 10000);
+        require(_pct_bps < BPS_ONE_HUNDRED_PCT); // DssExecLib/cut-too-high
+        setValue(_calc, "cut", rdiv(_pct_bps, BPS_ONE_HUNDRED_PCT));
     }
 
     /*************************/
     /*** Oracle Management ***/
     /*************************/
+    /**
+        @dev Allows an oracle to read prices from it's source feeds
```
e18r, line 818: ~~it's~~ its

src/DssExecLib.sol
```diff
@@ -779,64 +807,82 @@ library DssExecLib {
...
+        @param _oracle  An OSM or LP oracle contract
+    */
+    function whitelistOracleMedians(address _oracle) public {
+        address median0;
+        (bool ok, bytes memory data) = _oracle.call(abi.encodeWithSignature("orb0()"));
+        if (ok) {
+            // Token is an LP oracle
+            median0 = abi.decode(data, (address));
```
e18r, line 826:
```suggestion
        (bool ok, bytes memory data) = _oracle.call(abi.encodeWithSignature("orb0()"));
        if (ok) {
            // Token is an LP oracle
            address median0 = abi.decode(data, (address));
```

src/test/DssAction.t.sol
```diff
@@ -734,6 +741,32 @@ contract ActionTest is DSTest {
...
+        assertEq(median.bud(tokenPip), 0);
+        action.whitelistOracleMedians_test(tokenPip);
+        assertEq(median.bud(tokenPip), 1);
+    }
+
+    function test_whitelistOracle_LP() public {
+        // Mock an LP oracle and whitelist it
+        address token0 = address(new DSToken("nil"));
+        address token1 = address(new DSToken("one"));
+        Median  med0   = new Median();
```
e18r, line 756: In real life this would be the medianizers relying the Pause Proxy, right?

src/test/DssAction.t.sol
```diff
@@ -734,6 +741,32 @@ contract ActionTest is DSTest {
...
+        Median  med1   = new Median();
+        address lperc  = address(new UniPairMock(token0, token1));
+        med0.rely(address(action));
```
e18r, line 759: In real life this would be the medianizers relying the Pause Proxy, right?

test.sh
```diff
@@ -6,4 +6,11 @@ set -e
 # shellcheck disable=SC1091
 source "./allow-optimize.sh"
 
-DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200 dapp --use solc:0.6.12 test -v --rpc-url="$ETH_RPC_URL"
+export DAPP_BUILD_OPTIMIZE=1
```
e18r, line 9: I just checked build.sh and we are also using the optimizer. Is this really what we want? This is pretty critical code.

e18r (commented):
LGTM except for the optimizer. I think it should be disabled.

gbalabasquer:
> LGTM except for the optimizer. I think it should be disabled.

IMO for this one can be enabled. It is not a core and permanent contract. The risks a bug on the optimizer

e18r:
After further discussions with the team, the optimizer is a separate issue. I'm OK with merging this PR without taking into account the optimizer.

src/test/DssAction.t.sol
brianmcmichael: Yes. The DssAction test file is only testing that the actions are updating the contract values correctly using a test deployment. RPC tests against the pauseProxy are happening in DssExec

test.sh
brianmcmichael: AFAIK @gbalabasquer has been deploying exec's optimized for 1 run. We should confirm whether we're still doing that.

brianmcmichael:
I think I've addressed all comments, ready for another look @gbalabasquer @e18r @hexonaut

PR #79, opened 2021-07-15
Make description required in DssAction

brianmcmichael:
Moves the description to a required implementation of DssAction. 
This means that it's no longer necessary to pass the description as a parameter of DssExec.
Saves some gas, and fails compilation if it's missing from the spell, now.


kmbarry1 (approved):
LGTM

kmbarry1:
I'm going to go ahead and merge this so the spell is maximally ready for review tomorrow AM.

gbalabasquer (commented):
Just to give an extra approval. All looks good to me as well.

PR #80, opened 2021-07-21
Test lerp ilk and upd readme

brianmcmichael:
Adds a test of the ilk-specific `linearInterpolation` function and updates the readme to include lerp library calls.


hexonaut (approved):
LGTM 👍

PR #81, opened 2021-08-20
Fix LP oracle whitelisting and clean up seldom used oracle functions

brianmcmichael:
* Fixes a bug where a UNIV2 token onboarding with a DSValue as one of the price sources will fail.
* Simplifies the oracle functions by removing differentiation between OSM and Median
* Removes feed array addition of oracle readers from the exec lib.

### DIFF
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -833,61 +833,44 @@ library DssExecLib {
         if (ok) {
             // Token is an LP oracle
             address median0 = abi.decode(data, (address));
-            addReaderToMedianWhitelist(median0, _oracle);
-            addReaderToMedianWhitelist(OracleLike(_oracle).orb1(), _oracle);
+            addReaderToWhitelistCall(median0, _oracle);
+            addReaderToWhitelistCall(OracleLike(_oracle).orb1(), _oracle);
         } else {
             // Standard OSM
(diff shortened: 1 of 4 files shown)


src/DssExecLib.sol
```diff
@@ -833,61 +833,44 @@ library DssExecLib {
...
             // Token is an LP oracle
             address median0 = abi.decode(data, (address));
-            addReaderToMedianWhitelist(median0, _oracle);
-            addReaderToMedianWhitelist(OracleLike(_oracle).orb1(), _oracle);
+            addReaderToWhitelistCall(median0, _oracle);
+            addReaderToWhitelistCall(OracleLike(_oracle).orb1(), _oracle);
         } else {
             // Standard OSM
-            addReaderToMedianWhitelist(OracleLike(_oracle).src(), _oracle);
+            addReaderToWhitelistCall(OracleLike(_oracle).src(), _oracle);
```
gbalabasquer, line 840: If it is an standard OSM, then we could use the regular function.

gbalabasquer: Nvm, can be a standard `OSM` but still sourcing from a `DSValue`.

gbalabasquer (approved):
LGTM

PR #83, opened 2021-10-04
Hole gt dust

brianmcmichael:
* Update to ensure ilk.hole >= ilk.dust
* Remove old test requirement
* Change order of operations in collateral onboarding due to new requirement


src/DssExecLib.sol
```diff
@@ -662,6 +666,8 @@ library DssExecLib {
     */
     function setIlkMinVaultAmount(bytes32 _ilk, uint256 _amount) public {
         require(_amount < WAD);  // "LibDssExec/incorrect-ilk-dust-precision"
+        (,, uint256 _hole,) = DogLike(dog()).ilks(_ilk);
+        require(_amount <= _hole / RAD);  // Ensure ilk.hole >= dust
         setValue(vat(), _ilk, "dust", _amount * RAD);
```
gbalabasquer, line 671: We might want to cache in a variable the amt in RAD and use it in the require and then in the `setValue`. Btw any reason why not using safe math here?

gbalabasquer (approved):
LGTM

PR #86, opened 2021-11-24
Add collateral offboarding

iamchrissmith:
Seeing several spells about off boarding collateral, I wanted to try adding an Exec Lib function that would handle it for us.

### DIFF
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -1066,19 +1066,21 @@ library DssExecLib {
 
     /**
         @dev Performs basic functions and sanity checks to start removing a collateral type from the MCD system
-        @param _ilk      Collateral type key code [Ex. "ETH-A"]
-        @param _lerp     Name to be used for the LERP [Ex. "ETH-A Offboarding"]
-        @param _matCur   Start value for the mat LERP
-        @param _matEnd   End value for the mat LERP
-        @param _duration Duration of the mat LERP interpolation
-        @return line     current line value for the ilk to be used to reduce the total Line
+        @param _ilk         Collateral type key code [Ex. "ETH-A"]
+        @param _lerp        Name to be used for the LERP [Ex. "ETH-A Offboarding"]
+        @param _matCur      Start value for the mat LERP
+        @param _matEnd      End value for the mat LERP
+        @param _duration    Duration of the mat LERP interpolation
+        @param _reduceLine  if true will also reduce the global Line
+        @return line        ilk line value for the ilk
     */
     function startOffboardingCollateral(
         bytes32 _ilk,
         bytes32 _lerp,
         uint256 _matCur,
         uint256 _matEnd,
-        uint256 _duration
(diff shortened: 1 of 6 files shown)


src/DssExecLib.sol
```diff
@@ -1037,6 +1037,47 @@ library DssExecLib {
...
+        @param _duration Duration of the mat LERP interpolation
+        @return line     current line value for the ilk to be used to reduce the total Line
+    */
+    function startOffboardingCollateral(
+        bytes32 _ilk,
+        bytes32 _lerp,
+        uint256 _matCur,
+        uint256 _matEnd,
+        uint256 _duration
+    ) public returns (uint256 line) {
```
naszam, line 1059: I would adjust the `Line` inside the action itself to avoid adding extra logic in the spell or forgetting about it.

brianmcmichael: Could probably grab the current ilk.line out of the system and then use the `decreaseIlkDebtCeiling(bytes32 _ilk, uint256 _amount, bool _global)` on Line 1068 to handle the Ilk `line` and Global `Line` in one call.

iamchrissmith: I decided to parametrize that decision so that if we are offboarding multiple collaterals we can only call `decreaseGlobalLine` once with the total decrease, what do you think of that?

brianmcmichael: That should be documented pretty well if so. TBH I'd rather see these functions be fairly self-contained to avoid "gotcha"s in the implementation. Maybe a "global" `bool` parameter here like in the decrease function that could skip it if false?

And tbh I still don't know if governance knows what "offboarding" a collateral looks like yet so I wonder if it's too early for this. We may not stick with this model as there have been some complaints that it's catching people by surprise.

src/DssExecLib.sol
```diff
@@ -1065,7 +1067,7 @@ library DssExecLib {
 
         setIlkLiquidationPenalty(_ilk, 0);
         removeIlkFromAutoLine(_ilk);
-        setIlkDebtCeiling(_ilk, 0);
+        decreaseIlkDebtCeiling(_ilk, line / RAD, _reduceLine);
```
brianmcmichael, line 1070: I'm not sure this will work, unfortunately. It will leave a little bit of dust due to the rounding.

We might need to stick with setting the debt ceiling to zero and then having an `if` check on `_reduceLine` that subtracts the exact value from the global line using a direct vat call.

iamchrissmith: good call! I updated it to a direct call

naszam:
we might need to `dapp upgrade dss-interfaces` in https://github.com/makerdao/dss-direct-deposit and here as well to sync with https://github.com/makerdao/dss-interfaces/pull/64

iamchrissmith:
closing for now.  We want to see how offboarding does or does not evolve.  Can refer back to this during the next offboarding to see if it is helpful

PR #89, opened 2021-11-30
D3m bar

brianmcmichael:
Adds function for setting D3M `bar` to the DssExecLib

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,9 @@ Below is an outline of all functions used in the library.
 - `setMedianWritersQuorum(address _median, uint256 _minQuorum)`: Sets the minimum number of valid messages from whitelisted oracle feeds needed to update median price.
 - `allowOSMFreeze(address _osm, bytes32 _ilk)`: Add OSM address to OSM mom, allowing it to be frozen by governance.
 
+### Direct Deposit Module
+- `setD3MTargetInterestRate(address _d3m, uint256 _pct_bps)`: Set the target rate (`bar`) for the Aave D3M module.
+
 ### Collateral Onboarding
 In order to onboard new collateral to the Maker protocol, the following must be done before the spell is prepared:
 - Deploy a GemJoin contract
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ Below is an outline of all functions used in the library.
 - `allowOSMFreeze(address _osm, bytes32 _ilk)`: Add OSM address to OSM mom, allowing it to be frozen by governance.
 
 ### Direct Deposit Module
-- `setD3MTargetInterestRate(address _d3m, uint256 _pct_bps)`: Set the target rate (`bar`) for the Aave D3M module.
+- `setD3MTargetInterestRate(address _d3m, uint256 _pct_bps)`: Set the target rate (`bar`) for a D3M module.
 
 ### Collateral Onboarding
 In order to onboard new collateral to the Maker protocol, the following must be done before the spell is prepared:
--- a/src/DssExecLib.sol
+++ b/src/DssExecLib.sol
@@ -902,7 +902,7 @@ library DssExecLib {
     /**
         @dev Sets the target rate threshold for a dai direct deposit module (d3m)
         @dev Aave: Targets the variable borrow rate
-        @param _d3m     The address of the ExponentialDecrease pricing contract
+        @param _d3m     The address of the D3M contract
         @param _pct_bps Target rate in basis points. (ex. 4% == 400)
     */
     function setD3MTargetInterestRate(address _d3m, uint256 _pct_bps) public {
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -174,6 +174,9 @@ Below is an outline of all functions used in the library.
 - `setMedianWritersQuorum(address _median, uint256 _minQuorum)`: Sets the minimum number of valid messages from whitelisted oracle feeds needed to update median price.
 - `allowOSMFreeze(address _osm, bytes32 _ilk)`: Add OSM address to OSM mom, allowing it to be frozen by governance.
 
+### Direct Deposit Module
+- `setD3MTargetInterestRate(address _d3m, uint256 _pct_bps)`: Set the target rate (`bar`) for the Aave D3M module.
```
hexonaut, line 178: This design will probably be generalized, so I think we don't have to mention "Aave".

src/DssExecLib.sol
```diff
@@ -895,6 +895,20 @@ library DssExecLib {
         MomLike(osmMom()).setOsm(_ilk, _osm);
     }
 
+    /*****************************/
+    /*** Direct Deposit Module ***/
+    /*****************************/
+
+    /**
+        @dev Sets the target rate threshold for a dai direct deposit module (d3m)
+        @dev Aave: Targets the variable borrow rate
```
hexonaut, line 904: Same here. Let's leave this generic to both Aave/Compound and whoever else in the future.

hexonaut (commented):
Two minor changes, but otherwise looks good.

README.md
brianmcmichael: Yes, I had to write the tests to be aave-specific, but I suppose there's nothing in the call here that needs to be. I'd love if we can standardize this.

src/test/DssAction.t.sol
```diff
@@ -76,6 +78,20 @@ contract UniPairMock {
     }
 }
 
+contract AaveMock {
+    // https://docs.aave.com/developers/the-core-protocol/lendingpool
+    function getReserveData(address dai) public returns (
```
naszam, line 83: nit: the function can be restricted to `view` to silence warnings

src/test/DssAction.t.sol
```diff
@@ -76,6 +78,20 @@ contract UniPairMock {
...
+        uint256, uint128, uint128, uint128, uint128, uint128, uint40, address, address, address, address, uint8
+    ) {
+        address _dai = dai; // avoid stack too deep
+        return (0,0,0,0,0,0,0, _dai, _dai, _dai, address(this), 0);
+    }
+
+    function getMaxVariableBorrowRate() public returns (uint256) {
```
naszam, line 90: nit: the function can be restricted to `pure` to silence warnings

src/DssExecLib.sol
brianmcmichael: I think we should add a line for Compound later. This is to specifically point out that for Aave, we're targeting the borrow rate (and not the lending rate). Every protocol is going to be different.

README.md
brianmcmichael: a069c02aa2b3f20a8d6705a6e251096a79790e09

src/DssExecLib.sol
hexonaut: I suspect it will be the same. Really we could pick either in every lending market, but we are interested in the borrow rate as that is what is competitive with Maker.

hexonaut: I'm okay if you want to add a comment for every D3M type though.

src/test/DssAction.t.sol
brianmcmichael: 005429677088e6492457ba8ed216fd39d9854e73

src/test/DssAction.t.sol
brianmcmichael: 005429677088e6492457ba8ed216fd39d9854e73

naszam (commented):
Checked:

* `lib/dss-direct-deposit` points to [`773069`](https://github.com/makerdao/dss-direct-deposit/tree/773069137357080d666543b040217b9a45c60315) (not latest on master)
* README Changes Look Good
* DssExecLib D3M Action Looks Good 
* Aave Mock Looks Good
* Tests Look Good
* Tests PASS

nit: noticed a recent commit on the d3m repo, we could consider to `dapp upgrade dss-direct-deposit` to latest on master and eventually consider to release a new d3m version @hexonaut (as some minor changes in code have been introduced since the first release) and point to the release tag.

LGTM :+1:

hexonaut (approved):
LGTM. Noted about the dss-direct-deposit version @naszam

naszam (approved):
LGTM

brianmcmichael:
Latest master on D3M is just a little change I made to README. No functional difference.

PR #92, opened 2022-02-02
dapp upgrade dss-interfaces

naszam:
sync with https://github.com/makerdao/dss-interfaces/pull/64


brianmcmichael:
Can you upgrade again? There was another update to dss-interfaces this morning.

brianmcmichael (approved):

Issue #42, opened 2021-01-15
All visibility modifiers on DssAction should be internal


hexonaut:
It doesn't look like any of the functions in DssActions are used outside of the inheriting SpellAction. You could get some decent byte savings by setting the visibility to internal on everything. That way it only compiles if it's referenced.

brianmcmichael:
This is planned for early next week.

hexonaut:
Looks like this has been done. Closing.

PR #93, opened 2022-03-01
set RWA DC

julienmartinlevrai:


godsflaw (approved):
LGTM

src/DssAction.sol
```diff
@@ -2,7 +2,7 @@
 //
 // DssAction.sol -- DSS Executive Spell Actions
 //
-// Copyright (C) 2020 Maker Ecosystem Growth Holdings, Inc.
+// Copyright (C) 2018-2022 Dai Foundation
```
brianmcmichael, line 5: I'm not sure this is appropriate. The Dai Foundation and this code did not exist in 2018 and this code in particular was developed under the Maker Foundation.
I think it's probably more appropriate to keep the original line and add a new one reflecting the Dai Foundation's copyright claim.

src/DssAction.sol
```diff
@@ -59,7 +59,7 @@ abstract contract DssAction {
     function description() external virtual view returns (string memory);
 
     // Returns the next available cast time
-    function nextCastTime(uint256 eta) external returns (uint256 castTime) {
+    function nextCastTime(uint256 eta) external view returns (uint256 castTime) {
```
brianmcmichael, line 62: Similar comment to line #42.

src/DssAction.sol
```diff
@@ -39,7 +39,7 @@ abstract contract DssAction {
     // Office Hours defaults to true by default.
     //   To disable office hours, override this function and
     //    return false in the inherited action.
-    function officeHours() public virtual returns (bool) {
+    function officeHours() public virtual pure returns (bool) {
```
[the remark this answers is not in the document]
julienmartinlevrai: Some sort of conditional office hours ?

src/DssAction.sol
julienmartinlevrai: What do you mean ? This function is not virtual.

src/DssAction.sol
julienmartinlevrai: pushed

src/DssAction.sol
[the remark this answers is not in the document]
julienmartinlevrai: I tested a non-pure override, and it causes a compile-time error. So I will change this to `view`, since I think we can agree that this function does not need to modify state.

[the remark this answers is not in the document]
julienmartinlevrai: pushed

src/DssExecLib.sol
```diff
@@ -638,6 +643,18 @@ library DssExecLib {
         setValue(_vat, _ilk, "line", sub(line_, _amount * RAD));
         if (_global) { decreaseGlobalDebtCeiling(_amount); }
     }
+    /**
+        @dev Set a RWA collateral debt ceiling by specifying its new oracle price.
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
+        @param _price    The new oracle price in natural units
```
talbaneth, line 650: Just a suggestion - since `_price` is a tricky parameter to configure, maybe add a comment hinting on how to calculate it. Something like:
"_price should enable DAI to be drawn over the loan period while taking into account the configured ink amount, interest rate and liquidation ratio".
See example in https://github.com/makerdao/spells-mainnet/blob/master/archive/2021-07-17-DssSpell/DssSpell.sol#L53

julienmartinlevrai:
@naszam fixed, thank you

src/DssExecLib.sol
julienmartinlevrai: good suggestion, done

naszam:
Following https://github.com/makerdao/dss-direct-deposit/pull/35, I think we can `dapp upgrade dss-direct-deposit` dependency to reflect those changes but checking locally and with @gbalabasquer  we might need to remove the `lib/dss-interfaces` directory in `lib/dss-direct-deposit` manually as it seems it's not removed as part of the upgrade command:
> warning: unable to rmdir 'lib/dss-interfaces': Directory not empty

src/DssExecLib.sol
```diff
@@ -638,6 +643,20 @@ library DssExecLib {
...
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+    */
+    function setRWADebtCeiling(bytes32 _ilk, uint256 _ceiling, uint256 _price) public {
+        require(_ceiling < WAD);
+        require(_price < WAD);
+        setValue(vat(), _ilk, "line", _ceiling * RAD);
+        RwaOracleLike(getChangelogAddress("MIP21_LIQUIDATION_ORACLE")).bump(_ilk, _price * WAD);
```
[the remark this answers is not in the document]
julienmartinlevrai: done

julienmartinlevrai:
> Following [makerdao/dss-direct-deposit#35](https://github.com/makerdao/dss-direct-deposit/pull/35), I think we can `dapp upgrade dss-direct-deposit` dependency to reflect those changes but checking locally and with @gbalabasquer we might need to remove the `lib/dss-interfaces` directory in `lib/dss-direct-deposit` manually as it seems it's not removed as part of the upgrade command:
> 
> > warning: unable to rmdir 'lib/dss-interfaces': Directory not empty

done

talbaneth (approved):
LGTM

src/DssExecLib.sol
```diff
@@ -638,6 +643,21 @@ library DssExecLib {
...
     }
+    /**
+        @dev Set a RWA collateral debt ceiling by specifying its new oracle price.
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
+        @param _price    The new oracle price in natural units
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+    */
+    function setRWADebtCeiling(bytes32 _ilk, uint256 _ceiling, uint256 _price) public {
```
naszam, line 654: nit: consider naming the action `setRWAIlkDebtCeiling` following existing naming patterns.

julienmartinlevrai: done

src/DssExecLib.sol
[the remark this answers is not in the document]
brianmcmichael: I think it's probably ok to use the existing function here as it will inherit any guard rails we add later, though I'm not sure what guard rails we'd add to that function...

src/DssExecLib.sol
```diff
@@ -638,6 +643,21 @@ library DssExecLib {
...
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
+        @param _price    The new oracle price in natural units
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+    */
+    function setRWAIlkDebtCeiling(bytes32 _ilk, uint256 _ceiling, uint256 _price) public {
+        require(_ceiling < WAD);
+        require(_price < WAD);
+        setValue(vat(), _ilk, "line", _ceiling * RAD);
```
naszam, line 657: checking `bump` more closely, I think it assumes an increase in price `val` and so in the ilk debt ceiling, based on the formula in the spell, as the price should be originally set during the onboarding via the `init` function and the debt ceiling set following current collateral onboarding patterns.
I'm wondering if the action should also include the global debt ceiling increase (via `_global` flag) and maybe renamed into `increaseRWAIlkDebtCeiling` as well for clarity, following existing pattern (e.g. `increaseIlkDebtCeiling`).
cc: @godsflaw @brianmcmichael

julienmartinlevrai: You are right that in practice, the debt ceiling of MIP21 collateral will only increase. However, the semantics of the `increaseIlkDebtCeiling()` function would require the `_ceiling` parameter to be a delta instead of an absolute amount. This is problematic for the `_price` parameter, because if it is left as an absolut amount it breaks the semantics and can be confusing, and if it is required to be a delta it becomes harder to calculate.

brianmcmichael: We can decrease the ilk debt ceiling, we just can't decrease the price.

`set` is therefore probably fine here, but maybe we want a require or at least a dev comment that `_price` is greater than or equal to the pip price.

The `setIlkDebtCeiling()` function doesn't currently have a global bool because it wasn't clear if the added overhead to determine whether an increase or decrease was happening was worth the expense.

brianmcmichael, line 657:
```suggestion
        setIlkDebtCeiling(_ilk, _ceiling);
```

julienmartinlevrai: I will add that as a comment as the require already is in place, except deeper in the call stack

julienmartinlevrai: done

src/DssExecLib.sol
```diff
@@ -638,6 +643,22 @@ library DssExecLib {
...
+        @dev Set a RWA collateral debt ceiling by specifying its new oracle price.
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
+        @param _price    The new oracle price in natural units
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+        @dev note: _price should be higher than the current oracle price
+    */
+    function setRWAIlkDebtCeiling(bytes32 _ilk, uint256 _ceiling, uint256 _price) public {
+        require(_ceiling < WAD);
```
naszam, line 656: I think this check can be removed as already present in [setIlkDebtCeiling](https://github.com/makerdao/dss-exec-lib/blob/master/src/DssExecLib.sol#L612)

src/DssExecLib.sol
```diff
@@ -638,6 +643,22 @@ library DssExecLib {
         setValue(_vat, _ilk, "line", sub(line_, _amount * RAD));
         if (_global) { decreaseGlobalDebtCeiling(_amount); }
     }
+    /**
+        @dev Set a RWA collateral debt ceiling by specifying its new oracle price.
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
```
naszam, line 649: I think we should keep the naming convention `_amount` to avoid confusion

julienmartinlevrai: Since the price is also an amount, I think it is actually less confusing to be more specific

src/DssExecLib.sol
julienmartinlevrai: done

src/DssExecLib.sol
naszam: I don't have a strong objection, but I think it would keep it consistent with the underlying `setIlkDebtCeiling` input parameter naming convention.

src/DssExecLib.sol
```diff
@@ -638,6 +643,21 @@ library DssExecLib {
...
+        @param _price    The new oracle price in natural units
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+        @dev note: _price should be higher than the current oracle price
```
naszam, line 653:
```suggestion
        @dev note: _price should be greater than or equal to the current oracle price
```

src/DssAction.sol
```diff
@@ -39,7 +40,7 @@ abstract contract DssAction {
     // Office Hours defaults to true by default.
     //   To disable office hours, override this function and
     //    return false in the inherited action.
-    function officeHours() public virtual returns (bool) {
+    function officeHours() public virtual view returns (bool) {
```
naszam, line 43: nit: I would put the `view` before the `virtual`

src/DssExec.sol
```diff
@@ -2,7 +2,8 @@
 //
 // DssExec.sol -- MakerDAO Executive Spell Template
 //
-// Copyright (C) 2020 Maker Ecosystem Growth Holdings, Inc.
+// Copyright (C) 2020-2021 Maker Ecosystem Growth Holdings, Inc.
+// Copyright (C) 2021-2022 Dai Foundation
```
naszam, line 6: wondering if the license changes are correct (dates and keep the old one together with the new one) in the source files, following the recent discussion on `dss` repo, or it should be only Dai Foundation?
cc: @gbalabasquer

src/test/DssTestAction.sol
```diff
@@ -31,7 +32,7 @@ contract DssTestNoOfficeHoursAction is DssAction {
         require(!officeHours());
     }
 
-    function officeHours() public override returns (bool) {
+    function officeHours() public override view returns (bool) {
```
naszam, line 35: nit: same as above, I would put the `view` after `public`

naszam:
Checked (deps):
- MIP21-RWA-Example (installed)
    - `69bb350` (latest on master)
- dss (upgraded, changes introduced in `join.sol` (events)
    - `c666ab1` (not latest on master)
- dss-direct-deposit (upgraded)
    - `f1e01ad` (latest on master)
- dss-interfaces (upgraded)
    - `a82bc3a` (not latest on master)
- ilk-registry (upgraded)
    - `81ba973` (latest on master)
  
I'm not sure if more dependencies upgrades are required at this stage and regarding the `dss` upgrade, I'm wondering if it's required and if it should be kept in sync here if there are no remapping issues (It seems that last upgrade was pointing at `d688ebb` due to `dss-test`, `dss-token` and `ds-value` upgrades).
  cc: @brianmcmichael

src/DssExecLib.sol
```diff
@@ -638,6 +643,21 @@ library DssExecLib {
...
+    /**
+        @dev Set a RWA collateral debt ceiling by specifying its new oracle price.
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
+        @param _price    The new oracle price in natural units
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+        @dev note: _price should be greater than or equal to the current oracle price
+    */
+    function setRWAIlkDebtCeiling(bytes32 _ilk, uint256 _ceiling, uint256 _price) public {
```
naszam, line 655: Consider adding documentation for the action in the readme regarding the behaviour and usage.

src/DssAction.sol
julienmartinlevrai: done

src/DssExec.sol
julienmartinlevrai: done

julienmartinlevrai:
> I'm not sure if more dependencies upgrades are required at this stage and regarding the `dss` upgrade, I'm wondering if it's required and if it should be kept in sync here if there are no remapping issues (It seems that last upgrade was pointing at `d688ebb` due to `dss-test`, `dss-token` and `ds-value` upgrades). cc: @brianmcmichael

I think it is a good idea to upgrade. I don’t know in what dependency hell I am getting myself into, but yes.

julienmartinlevrai:
> > I'm not sure if more dependencies upgrades are required at this stage and regarding the `dss` upgrade, I'm wondering if it's required and if it should be kept in sync here if there are no remapping issues (It seems that last upgrade was pointing at `d688ebb` due to `dss-test`, `dss-token` and `ds-value` upgrades). cc: @brianmcmichael
> 
> I think it is a good idea to upgrade. I don’t know in what dependency hell I am getting myself into, but yes.

After talking with @naszam, I agree that this dependencies upgrade deserves its own set of PRs.

src/DssExecLib.sol
naszam: wondering if it would be better to include the action into a separate action sub section.

src/test/DssAction.t.sol
```diff
@@ -233,6 +241,36 @@ contract ActionTest is DSTest {
...
+        uint256 mat,
+        address operator
+    ) internal {
+        uint256 val = line / RAY; // TODO calculation
+        rwaOracle.init(ilk, val, doc, tau);
+        (,address pip,,) = rwaOracle.ilks(ilk);
+        spot.file(ilk, "pip", pip);
+        vat.init(ilk);
+        jug.init(ilk);
+        DSToken token = new DSToken(string(abi.encodePacked(ilk)));
```
naszam, line 258: wondering why the `RwaToken` is not used instead.

src/test/DssAction.t.sol
```diff
@@ -233,6 +241,36 @@ contract ActionTest is DSTest {
...
 
+    function init_rwa(
+        bytes32 ilk,
+        uint256 line,
+        uint48 tau,
+        uint256 duty,
+        uint256 mat,
+        address operator
+    ) internal {
+        uint256 val = line / RAY; // TODO calculation
```
naszam, line 252: It seems the `val` (`price`) calculation is missing

src/test/DssAction.t.sol
```diff
@@ -651,6 +729,15 @@ contract ActionTest is DSTest {
         action.decreaseIlkDebtCeiling_test("gold", 101 * MILLION); // Fail
     }
 
+    function test_setRWAIlkDebtCeiling() public {
+        action.setRWAIlkDebtCeiling_test("6s", 50 * MILLION, 55 * MILLION);
+        (,,, uint256 line,) = vat.ilks("6s");
+        assertEq(line, 50 * MILLION * RAD);
+        (,address pip,,) = rwaOracle.ilks("6s");
+        uint256 price = uint256(PipLike(pip).read());
+        assertEq(price, 55 * MILLION * WAD);
```
naszam, line 738: I would include a test to validate the `val` formula in the `init_rwa` against the [computation](https://github.com/makerdao/spells-mainnet/blob/master/archive/2021-07-17-DssSpell/DssSpell.sol#L55) showed in the spell, using the [assertEqApprox](https://github.com/makerdao/spells-mainnet/blob/master/src/DssSpell.t.base.sol#L145) test assertion.

```suggestion
        (,address pip,,) = rwaOracle.ilks("6s");
        uint256 price = uint256(PipLike(pip).read());
        assertEqApprox(price, 22_278_900 * WAD, WAD); // 20MM * 1.03^2 * 1.05
        action.setRWAIlkDebtCeiling_test("6s", 50 * MILLION, 55 * MILLION);
        (,,, uint256 line,) = vat.ilks("6s");
        assertEq(line, 50 * MILLION * RAD);
        price = uint256(PipLike(pip).read());
        assertEq(price, 55 * MILLION * WAD);
```

src/test/DssAction.t.sol
```diff
@@ -651,6 +729,17 @@ contract ActionTest is DSTest {
         action.decreaseIlkDebtCeiling_test("gold", 101 * MILLION); // Fail
     }
 
+    function test_setRWAIlkDebtCeiling() public {
+        (,address pip,,) = rwaOracle.ilks("6s");
+        uint256 price = uint256(PipLike(pip).read());
+        assertEqApprox(price, 22_278_900 * WAD, WAD); // 20MM * 1.03^2 * 1.05
```
naszam, line 735: `assertEqApprox` is required as not provided by `ds-test`

src/test/DssAction.t.sol
julienmartinlevrai: done

src/test/DssAction.t.sol
julienmartinlevrai: added

src/test/DssAction.t.sol
```diff
@@ -651,6 +746,17 @@ contract ActionTest is DSTest {
...
+        action.setRWAIlkDebtCeiling_test("6s", 50 * MILLION, 55 * MILLION);
+        (,,, uint256 line,) = vat.ilks("6s");
+        assertEq(line, 50 * MILLION * RAD);
+        price = uint256(PipLike(pip).read());
+        assertEq(price, 55 * MILLION * WAD);
+    }
+
```
naszam, line 759: I would include a fail test for the edge case where `_price` < `pip.read` (current oracle price), following previous discussion.

```solidity
    function testFail_setRWAIlkDebtCeiling() public {
        action.setRWAIlkDebtCeiling_test("6s", 50 * MILLION, 20 * MILLION); // Fail
    }
```

src/test/DssAction.t.sol
```diff
@@ -651,6 +746,17 @@ contract ActionTest is DSTest {
...
 
+    function test_setRWAIlkDebtCeiling() public {
+        (,address pip,,) = rwaOracle.ilks("6s");
+        uint256 price = uint256(PipLike(pip).read());
+        assertEqApprox(price, 22_278_900 * WAD, WAD); // 20MM * 1.03^2 * 1.05
+        action.setRWAIlkDebtCeiling_test("6s", 50 * MILLION, 55 * MILLION);
+        (,,, uint256 line,) = vat.ilks("6s");
+        assertEq(line, 50 * MILLION * RAD);
+        price = uint256(PipLike(pip).read());
+        assertEq(price, 55 * MILLION * WAD);
```
naszam, line 757: Consider extend the test to also include ceiling decrease, where the `_price` is equal to `pip.read` (current oracle price) following previous discussion.

```suggestion
        action.setRWAIlkDebtCeiling_test("6s", 50 * MILLION, 55 * MILLION); // Increase
        (,,, uint256 line,) = vat.ilks("6s");
        assertEq(line, 50 * MILLION * RAD);
        price = uint256(PipLike(pip).read());
        assertEq(price, 55 * MILLION * WAD);
        action.setRWAIlkDebtCeiling_test("6s", 40 * MILLION, 55 * MILLION); // Decrease
        (,,, line,) = vat.ilks("6s");
        assertEq(line, 40 * MILLION * RAD);
        price = uint256(PipLike(pip).read());
        assertEq(price, 55 * MILLION * WAD);
```

naszam (commented):
Checked:

- License Changes
- DssExecLib `setRWAIlkDebtCeiling` Action
- Action Test Coverage
- README Changes
- Test PASS

Notes: left some suggestions to improve test coverage

LGTM

src/test/DssAction.t.sol
julienmartinlevrai: done

naszam (approved):
LGTM

src/DssAction.sol
```diff
@@ -30,17 +30,19 @@ abstract contract DssAction {
 
     using DssExecLib for *;
 
+    // Office Hours defaults to true by default.
+    bool OFFICE_HOURS = true;
```
talbaneth, line 34: We shouldn't use storage variables in the spell action.
Specifically this field will collide with the `owner` field of the pause proxy 
during spell execution (so you should get wrong office hours value as it reads the delegating contract stoarge slot):

`https://github.com/dapphub/ds-pause/blob/master/src/pause.sol#L110`

src/DssExecLib.sol
```diff
@@ -638,6 +642,21 @@ library DssExecLib {
...
         if (_global) { decreaseGlobalDebtCeiling(_amount); }
     }
+    /**
+        @dev Set a RWA collateral debt ceiling by specifying its new oracle price.
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
+        @param _price    The new oracle price in natural units
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+        @dev note: _price should be greater than or equal to the current oracle price
```
talbaneth, line 652: Technically it's `_price * WAD` that should be greater or equal to the current oracle price.
We might want to change the comment.

src/DssAction.sol
julienmartinlevrai: This is a very good point, @talbaneth, thank you so much for spotting that !
This takes us back to the discussion with @brianmcmichael about the state mutability modifier of the `officeHours` function. Since that function will be executed in the context of the Pause Proxy, it does not make sense for it to be a `view` because the only storage slot it could read is `owner`. So I will change it to `pure`.

julienmartinlevrai: done

src/DssExecLib.sol
julienmartinlevrai: done

src/DssAction.sol
talbaneth: mm we might still want to do external calls to non pure functions from other contracts.
I think it's ok to leave it as before for now and ignore the warning.

julienmartinlevrai: done

src/DssExecLib.sol
```diff
@@ -638,6 +643,21 @@ library DssExecLib {
...
         if (_global) { decreaseGlobalDebtCeiling(_amount); }
     }
+    /**
+        @dev Set a RWA collateral debt ceiling by specifying its new oracle price.
+        @param _ilk      The ilk to update (ex. bytes32("ETH-A"))
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
+        @param _price    The new oracle price in natural units
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+        @dev note: _price * WAD should be greater than or equal to the current oracle price
```
talbaneth, line 653:
```suggestion
        @dev note: _price * WAD should be greater than or equal to the current oracle price
```
I'm seeing that line with an `NBSP` special character:
image

I suspect this is maybe because you did the space when you were on french mode? (just guessing based on this - https://intellij-support.jetbrains.com/hc/en-us/community/posts/360009442799-2020-2-showing-NBSP-).

Let's see if this suggestion solves that.

talbaneth (approved):
Apart from the comment on the `NBSP` thing this looks good to me.

Would love to hear others opinion about this but in general I think it's better to do unrelated changes (adding `view` modifiers, license stuff) in separate PR/s.
Otherwise it drives focus away from the actual change and makes the reviewing harder.

julienmartinlevrai:
Hello @talbaneth I think you are right about not adding unrelated changes. Sorry for that, I will keep it in mind for future PRs.

talbaneth:
btw the `NPBS` thing is gone now, which is good.

src/DssExecLib.sol
```diff
@@ -638,6 +643,21 @@ library DssExecLib {
...
+        @param _ceiling  The new debt ceiling in natural units (e.g. set 10m DAI as 10_000_000)
+        @param _price    The new oracle price in natural units
+        @dev note: _price should enable DAI to be drawn over the loan period while taking into
+                   account the configured ink amount, interest rate and liquidation ratio
+        @dev note: _price * WAD should be greater than or equal to the current oracle price
+    */
+    function setRWAIlkDebtCeiling(bytes32 _ilk, uint256 _ceiling, uint256 _price) public {
+        require(_price < WAD);
+        setIlkDebtCeiling(_ilk, _ceiling);
+        RwaOracleLike(getChangelogAddress("MIP21_LIQUIDATION_ORACLE")).bump(_ilk, _price * WAD);
```
brianmcmichael, line 658: Interesting point to note here that `bump` is a `WAD` in this liquidation oracle and a `RAD` in the vow. Nothing actionable but pointing it out. Might be worth adding a comment to the MIP21 repo as the precision is not indicated there.

brianmcmichael (approved):
Looks good. Only minor comments.

src/test/DssAction.t.sol
```diff
@@ -185,6 +220,23 @@ contract ActionTest is DSTest {
         }
     }
 
+    // not provided in DSTest
+    function assertEqApprox(uint256 _a, uint256 _b, uint256 _tolerance) internal {
```
brianmcmichael, line 224: This could probably use some documentation. It looks like _tolerance is a straight delta but we might want to mention what format this takes. (i.e. not a percentage)

brianmcmichael (changes requested):
A couple of minor things around documentation.

src/test/DssAction.t.sol
```diff
@@ -142,6 +150,7 @@ contract ActionTest is DSTest {
     uint256 constant public RAD      = 10 ** 45;
 
     uint256 constant START_TIME = 604411200;
+    string constant doc = "QmdmAUTU3sd9VkdfTZNQM6krc9jsKgF2pz7W1qvvfJo1xk";
```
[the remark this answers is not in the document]
julienmartinlevrai: done

src/test/DssAction.t.sol
julienmartinlevrai: done

PR #95, opened 2022-05-17
make office hours explicit

iamchrissmith:
We should make office hours always explicit in the spells.


gbalabasquer (approved):
I like this change to be honest. It forces that there is always an implementation of `officeHours` in the spell and doesn't assume anything as default.

PR #97, opened 2022-11-30
Remove dependencies and move from dapp to forge

gbalabasquer:


gbalabasquer:
Once this gets merged I'll move on with a new PR for upgrading to solc 0.8.x.

Makefile
```diff
@@ -1,4 +1,4 @@
-build        :; ./build.sh
+build        :; DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200 dapp --use solc:0.6.12 build
```
iamchrissmith, line 1: if we are moving from dapp to forge, do we want to switch these too:
```
forge build
forge clean
forge create
```
etc?

iamchrissmith (changes requested):
one clean up item and two questions

:mechanical_arm:

src/test/DssAction.t.sol
```diff
@@ -195,9 +505,9 @@ contract ActionTest is DSTest {
     function dai(address urn) internal view returns (uint) {
         return vat.dai(urn) / RAY;
     }
-    function gem(bytes32 ilk, address urn) internal view returns (uint) {
-        return vat.gem(ilk, urn);
-    }
+    // function gem(bytes32 ilk, address urn) internal view returns (uint) {
+    //     return vat.gem(ilk, urn);
+    // }
```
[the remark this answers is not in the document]
gbalabasquer: Done!

src/test/DssAction.t.sol
```diff
@@ -20,138 +20,448 @@
...
+interface Univ2OracleFactoryLike {
+    function build(address, address, bytes32, address, address) external returns (address oracle);
+}
+
+interface D3MLike {
+    function bar() external view returns (uint256);
+    function rely(address) external;
+}
+
+contract MockUniPair {
```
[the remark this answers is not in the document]
gbalabasquer: Done!

Makefile
gbalabasquer: the thing is `forge create` apparently doesn't work the same than `dapp create` (we are still using the last one I believe). And we need to `build` using `dapp` in order to be able to create with it.
I'd guess `dapp clean` could be replaced without any problem.

gbalabasquer: Yeah actually, `forge clean` also removed `cache` directory which should be fine. For now I'm changing only that one.

gbalabasquer: Done!

iamchrissmith (approved):
lgtm!

rockyfour (approved):
Changes LGTM.
Tests pass.

src/test/Mocks.sol
```diff
@@ -0,0 +1,366 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
```
naszam, line 1:
```suggestion
// SPDX-FileCopyrightText: © 2022 Dai Foundation <www.daifoundation.org>
// SPDX-License-Identifier: AGPL-3.0-or-later
```

src/test/Mocks.sol
```diff
@@ -0,0 +1,366 @@
...
+//
+// Mocks.sol -- DSS Executive Spell Action Tests
+//
+// Copyright (C) 2022 Dai Foundation
```
naszam, line 5:
```suggestion
```

PR #98, opened 2022-12-05
Fix broken test

gbalabasquer:
This test was actually broken and not really proving the function was working (realized now that safe math inbuilt is running in my dev environment). I'll do it better in the solc 0.8 version but wanted to get it fixed before.


rockyfour (approved):
LGTM

